PR #12, opened 2026-05-04
feat(stt-v2): MInDS-14 corpus, stt-preview, docs, compose

seribaymadina:
## Summary

Adds the frozen **stt-v2** MInDS-14–style banking STT corpus (manifest + metadata), `coval-bench stt-preview` for no-DB probes against GCS audio, methodology and manifests documentation, local `docker-compose` stack for db/api/runner, and related runner fixes (blank GCS credentials, loader manifest fields, mypy for `datasets`).

## Test plan

- [ ] `cd runner && uv run pytest -q`
- [ ] Optional: `coval-bench stt-preview` with a provider key against public bucket

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Added STT v2 dataset (frozen 60‑clip conversational, phone-style US banking corpus) and CLI tools: stt-preview for provider sanity checks and publish/export commands for preparing/publishing reviewed STT manifests/WAVs.

* **Documentation**
  * Substantially expanded methodology and manifest docs with provenance, validation, normalization, versioning, and reproduction CLI workflow guidance.

* **Tests**
  * Added unit tests covering manifest generation, hashing, ordering, and metadata passthrough.


runner/src/coval_bench/stt_preview.py
```diff
@@ -0,0 +1,219 @@
...
+    for path in candidates:
+        if not path.is_file():
+            continue
+        try:
+            text = path.read_text(encoding="utf-8")
+        except OSError:
+            continue
+        merged.update(_parse_dotenv_text(text))
+    for key, val in merged.items():
+        os.environ.setdefault(key, val)
```
lorenjphillips, line 87: Merge order + `setdefault` means shell-exported vars always win over file values — the right behavior when someone has `DEEPGRAM_API_KEY` already set in their terminal for a different project and doesn't want a stale `.env.local` silently overriding it.

runner/src/coval_bench/stt_preview.py
```diff
@@ -0,0 +1,219 @@
...
+    items_out: list[dict[str, Any]] = []
+    all_ok = True
+
+    for item in dataset.items[:limit]:
+        row: dict[str, Any] = {"audio_filename": item.path.name, "path": str(item.path)}
+        pcm_blob = item.path.read_bytes()
+        dur_sec = item.duration_sec
+
+        try:
+            async with asyncio.timeout(_STT_TIMEOUT_S):
```
lorenjphillips, line 175: Timeout scoped to a single attempt with `with_retry` wrapping the outer call is the correct layering — wall-clock time is bounded per attempt, not per sequence. Swapping these would let a stuck provider consume arbitrarily long regardless of retry count.

runner/src/coval_bench/datasets/minds14_audio.py
```diff
@@ -0,0 +1,143 @@
...
+            "Install HF build deps: cd runner && uv sync --extra minds14-build "
+            "(or run with --extra minds14-build)"
+        ) from exc
+
+    rev = revision if revision is not None else hf_revision_env()
+    load_kw: dict[str, Any] = {"path": _HF_PATH, "name": _HF_NAME, "split": _HF_SPLIT}
+    if rev:
+        load_kw["revision"] = rev
+    ds = load_dataset(**load_kw)
+    return ds.cast_column("audio", Audio(decode=False))
```
lorenjphillips, line 71: Casting to `Audio(decode=False)` before iteration avoids pulling torchcodec/PyTorch into the runner's build footprint. The docstring explains the why clearly — keeping this decoding path in `soundfile` + `librosa` only is the right call for this environment.

docs/methodology.md
```diff
@@ -0,0 +1,178 @@
...
+A frozen STT corpus sourced from Hugging Face
+[`PolyAI/minds14`](https://huggingface.co/datasets/PolyAI/minds14) (`en-US`,
+`train`), **`CC-BY-4.0`**. The packaged manifest `stt-v2.json` is **authored and
+frozen**: each row keys off the Hub `path`, carries a human-reviewed reference
+transcript (and editorial metadata such as intent labels where present), and
+pins **`sha256`** for the mirrored WAV bytes. The JSON on disk remains the
+source of truth for transcripts and paths (reviewed text, not raw Hub
+captions).
+
+**Why this corpus (vs `stt-v1`).** `stt-v1` is clean read audiobook speech — ideal
```
lorenjphillips, line 30: This section is load-bearing — it's the first question anyone will ask when they see two leaderboards with very different WER numbers. Having the telephony bandwidth and call-center design rationale committed here instead of in a Slack thread means the context survives the next person who touches this dataset.

runner/tests/unit/test_publish_reviewed_stt_v2.py
```diff
@@ -0,0 +1,98 @@
...
+            "index": 0,
+            "approved_transcript": "hello",
+            "intent_class": 0,
+            "intent_name": "abroad",
+            "gender": "F",
+        },
+    }
+    d = _build_manifest_dict(review_by_path=review, wav_root=wav_dir)
+    item = d["items"][0]
+    assert item["intent_class"] == 0
```
lorenjphillips, line 94: Testing `intent_class == 0` explicitly matters — 0 is falsy, so a naive `if row.get('intent_class'):` guard in the publish script would silently drop it. Good edge case to pin in the test suite.

lorenjphillips (commented):
Solid work throughout — clean code, meaningful tests, and a methodology doc that will actually stay useful. Leaving a few notes on the highlights.

lorenjphillips (dismissed):

tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
```
lorenjphillips: This file shouldn't be at the repo root — it's a Next.js config that belongs inside `web/` if it's needed at all. Looks like a Cursor artifact. Drop it.

docs/methodology.md
```diff
@@ -0,0 +1,178 @@
...
+`config.py` next to the matrix entry.
+
+**Disabled entries.** Some entries are intentionally inert (`disabled=True`).
+The runner skips them; the API exposes them so that a frontend can grey out
+models that are catalogued but not currently producing data. See
+[ADR-011](#adr-references) for the policy.
+
+## 3. Normalization pipeline
+
+WER is computed after a deterministic normalization pipeline applied to both
```
lorenjphillips, line 138: This repo is public. The normalization pipeline details here — currency expansion, ordinals, date/time handling, unicode folding — hand providers a map of exactly what to tune for on our test set. Same concern applies to the dataset selection criteria and the full `stt-v2` intent breakdown above. Decision needed before merging: either move this file somewhere internal, or make a deliberate call that the transparency trade-off is worth it.

lorenjphillips (dismissed):
Two things to resolve before this merges — one is a clear drop, one needs a decision.

tsconfig.json
seribaymadina: Moved it to web/tsconfig.json where it belongs. It was needed as it provides the @/* path alias that every dashboard import relies on, and without it Vercel's build fails (confirmed: dropping the root copy without committing web/tsconfig.json broke the build). web/tsconfig.json was already on disk locally but had never been committed to git, so Vercel's clone didn't have it. Fixed it .

docs/methodology.md
seribaymadina: Decided to keep methodology.md public but strip the specifics. The pipeline details were removed. The new stt-v2 section documents the goals (duration band, intent breadth) and removed intent breakdown.

lorenjphillips (approved):
Both prior change requests addressed (root tsconfig dropped, normalization details moved internal). Architecture is clean, anonymous GCS reads work for OSS contributors, runner-rt SA handles Cloud Run, no new credentials needed. Rebasing onto current main before merge.

lorenjphillips (approved):
Re-approving rebased head — CI green across all 6 checks.

lorenjphillips:
Rebased onto current main and re-approved on the new head (`7352587`). Resolved my 7 affirmative/already-addressed threads.

Four substantive CodeRabbit threads still need your attention before this can merge:

1. `runner/src/coval_bench/datasets/manifests/stt-v2.json` lines 119/155/299 — transcript artifacts `servicesa and`, `balancea and`, `vacationa and` (stray `a` at clause boundaries). Will count as substitutions against every provider hypothesis and inflate WER artificially. Fix or confirm intentional.
2. `runner/src/coval_bench/datasets/scripts/export_minds14_reviewed_wavs.py:81` — `_load_hf_train` calls `load_dataset` without passing `HF_MINDS14_REVISION`, so re-running the export can pull a drifted HF snapshot whose decoded bytes no longer match `stt-v2.json` SHAs. Pin via the same env var that `minds14_audio.load_train_dataset` honors.
3. `runner/src/coval_bench/stt_preview.py` ~line 164-171 — `load_stt_dataset()` and `item.path.read_bytes()` are outside the guarded path, so dataset / bucket / object failures escape as unhandled exceptions instead of returning the structured JSON error payload the CLI promises.
4. `runner/tests/unit/test_orchestrator.py:155` — dataset stub returns `stt_dataset` for any non-`"tts-v1"` id, masking regressions where `settings.dataset_id` propagation breaks. Tighten to raise on unexpected ids.

Once those four are addressed and resolved, ruleset `required_review_thread_resolution` will let it merge.

PR #28, opened 2026-05-26
[BENCH-176] audit deepgram stt

coval-cale:
A few things came up:
1. `flux-general-multi` was using the wrong path when processing the transcript. Updated it.
2. Flux was hitting a preview endpoint instead of the production url.
3. Added `punctuate=true` for nova models. Should make WER comparisons fairer.
4. We were assuming a `SpeechEnded` event but it was really an `UtteranceEnd` event.

Also updated the tests a bit. Big one was adding happy path test to `flux-general-multi`.

## Summary by CodeRabbit

* **Improvements**
  * Deepgram speech-to-text now selects the correct protocol per model variant and produces more accurate, complete transcripts for flux and nova model families.
  * Message handling and transcript completion logic refined for more reliable real-time speech results.

* **Tests**
  * Test coverage tightened with stricter URL validation and exact transcript assertions, plus a new multi-stream happy-path scenario.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coval-ai/benchmarks/pull/28?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/runner/tests/providers/stt/test_deepgram.py
+++ b/runner/tests/providers/stt/test_deepgram.py
@@ -386,8 +386,7 @@ async def test_deepgram_flux_success(fake_api_key: SecretStr, audio_pcm_bytes: b
     assert result.ttft_seconds >= 0
     assert result.first_token_content is not None
     assert "hello" in result.first_token_content.lower()
-    assert result.complete_transcript is not None
-    assert "hello" in result.complete_transcript.lower()
+    assert result.complete_transcript == "hello world how are you"
     assert result.audio_to_final_seconds is not None
     assert result.audio_to_final_seconds >= 0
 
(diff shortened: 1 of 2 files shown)


runner/tests/providers/stt/test_deepgram.py
```diff
@@ -414,3 +414,34 @@ async def test_deepgram_flux_audio_to_final_none_when_no_transcript(
...
+            channels=1,
+            sample_width=2,
+            sample_rate=16000,
+            realtime_resolution=0.5,
+        )
+
+    assert result.error is None
+    assert result.ttft_seconds is not None
+    assert result.ttft_seconds >= 0
+    assert result.complete_transcript == "hello world how are you"
```
seribaymadina, line 445: fyi: this assertion would also pass under  the pre-fix code, since the fixture's last  TurnInfo happens to equal the longest. Not blocking,  Flux snapshots grow monotonically in practice, but flagging in case you ever want a fixture that truly disambiguates last-wins from longest-wins.

seribaymadina (approved):

runner/tests/providers/stt/test_deepgram.py
coval-cale: Totally. Just wanted to guard against the final transcript being shortened and this failing. Nice catch!

PR #32, opened 2026-05-27
[BENCH-180] Audit speechmatics stt implementation

coval-cale:
- Before: we were joining text which could lead to a higher WER due to improper punctuation parsing.
- After: Speechmatics provides a preformatted transcript string. Default to that, only parse if it isn't provided.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved speech-to-text transcript handling to prefer an explicit top-level transcript when present and reliably fall back to assembled results when absent or empty, improving accuracy and punctuation retention.

* **Tests**
  * Added unit tests and updated fixtures to validate transcript extraction across explicit, empty/whitespace, and missing transcript scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coval-ai/benchmarks/pull/32?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


lorenjphillips (approved):
Thank you!

PR #41, opened 2026-05-29
[BENCH-208] Update TTFA metric

coval-cale:
Digging into our metrics I noticed that we were seeing incredibly small variation in the TTFA (Time to First Audio) metric for some models, as low as std=10ms. Low variance is good but this picked my curiosity. I dug into how we measure TTFA and found that it was a simple difference between a T_0 representing when we send the transcript and a T_end when we receive the first chunk of audio file back. But the audio files we received back had significant differences in when the speech actually started. A provider could inadvertently inflate the TTFA metric by sending back a silent chunk of audio immediately after receiving the request.

This change adjusts how we compute TTFA. Before we computed time between sending the transcript and receiving the first audio. Now, we retain that but also process the audio to the time between the start of the audio and when the volume raises above a certain threshold (not exactly but an easy way to think about it). So our new TTFA is:

Time between sending the transcript and receiving the audio + time between audio start and audible speech

Also in this PR:
- Abstracted the wav file creation into a helper function. This was duplicated across every class so no significant functional change to the audio output.

This will lead to a slight inflation of provider latency compared to before. 

## Summary by CodeRabbit

* **Documentation**
  * Added TTFA methodology clarifying TTFA now measures perceived first‑audible latency (pre/post-change values not comparable) and updated section numbering.

* **Refactor**
  * Unified TTS output handling across providers to standardize timing capture, audio file creation behavior, and error handling (removes per-provider WAV-temp flows).

* **Tests**
  * Added unit tests for TTS finalization and TTFA calculations, plus minor test comment/docstring updates.

### DIFF
--- a/runner/src/coval_bench/providers/tts/_common.py
+++ b/runner/src/coval_bench/providers/tts/_common.py
@@ -1,0 +1,102 @@
+# Copyright 2026 The Coval Benchmarks Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared finalize step for TTS providers: perceived TTFA + WAV output."""
+
+from __future__ import annotations
+
+import os
+import tempfile
+import wave
+from pathlib import Path
+
+import structlog
+
+from coval_bench.metrics import compute_ttfa, first_audible_offset_ms
+from coval_bench.providers.base import TTSResult
+
+logger: structlog.BoundLogger = structlog.get_logger(__name__)
+
+# Mono signed 16-bit PCM: 2 bytes per sample.
+_PCM16_FRAME_BYTES = 2
+
+
+def finalize_tts_result(
+    *,
+    provider: str,
+    model: str,
+    voice: str,
+    pcm: bytes,
+    sample_rate: int,
+    audio_synthesis_start: float | None,
+    first_audio_chunk_at: float | None,
+    error: str | None = None,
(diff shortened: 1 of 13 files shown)


runner/src/coval_bench/providers/tts/_common.py
```diff
@@ -0,0 +1,100 @@
...
+    # Drop a dangling partial sample so neither offset detection (which rejects
+    # frame-misaligned PCM) nor the WAV header chokes on a split final frame.
+    remainder = len(pcm) % _PCM16_FRAME_BYTES
+    if remainder:
+        pcm = pcm[: len(pcm) - remainder]
+
+    ttfa_ms: float | None = None
+    if audio_synthesis_start is not None and first_audio_chunk_at is not None:
+        offset_ms = _safe_offset_ms(pcm, sample_rate, provider, model) if pcm else None
+        ttfa_ms = compute_ttfa(audio_synthesis_start, first_audio_chunk_at, offset_ms or 0.0)
```
lorenjphillips: minor nit — `offset_ms or 0.0` is fine since `0.0` is the only falsy value it can return, but `offset_ms if offset_ms is not None else 0.0` reads clearer next to the none-means-arrival-only contract

lorenjphillips (approved):

PR #42, opened 2026-05-29
Feat/http tts comparable ttfa

seribaymadina:
Align HTTP TTS benchmarking with the WebSocket methodology by reusing a warm HTTP/2 connection and excluding connection setup overhead from TTFA measurements.

changes made:
Apply HTTP/2, connection limits, and 300s keepalive in _http_session.py via TimedTransport.
Maintain a single warm HTTP/2 connection per host.
Add HTTP version logging and HTTP/1.1 fallback warnings in OpenAI and ElevenLabs warmup paths.
Surface warmup failures instead of swallowing them.
Scope warmup to the requested benchmark kind.
Update methodology documentation and add regression tests.

*_prewarm_no_http2 warns when HTTP/2 negotiation fails and the client falls back to HTTP/1.1.
This aligns HTTP benchmarking with the WebSocket cohort methodology.

## Summary by CodeRabbit

* **New Features**
  * Optional provider warmup phase to pre-warm HTTP connections; warmup failures are non-fatal and clients are shut down on completion

* **Documentation**
  * Added latency measurement convention clarifying TTFT/TTFA timing excludes protocol setup

* **Improvements**
  * Shared HTTP client pooling and HTTP diagnostics (version, setup timing, connection reuse) collected and persisted for improved TTS latency reporting

* **Tests**
  * Expanded HTTP, TTS provider, and orchestrator tests to cover pooling, timing stamps, warmup, and TTFA persistence

* **Chores**
  * Removed elevenlabs SDK dependency

### DIFF
--- a/docs/methodology.md
+++ b/docs/methodology.md
@@ -100,6 +100,64 @@ The DP edit-distance computation is delegated to `jiwer`. Pinned in
 `uv.lock` change is reviewed against the SPDX license-set policy before
 merge.
 
+## 5. Latency measurement convention
+
+TTFT (STT) and TTFA (TTS) measure model inference time. Pre-t0 deployment
+setup — TCP, TLS, protocol handshake, optional session-setup RTT — is
+excluded for every provider. The rule is applied uniformly across cohorts;
+the magnitude of what each cohort excludes varies by deployment architecture,
+but the rule is the same.
+
+| Cohort | What is excluded from t0 |
+|---|---|
+| WS streaming (Deepgram, AssemblyAI, ElevenLabs Scribe, Speechmatics, Gradium STT, Cartesia, Deepgram aura-2, Rime, Gradium TTS, Hume) | TLS + WS upgrade + optional session-setup RTT (~50–200 ms). Handshake naturally completes inside `measure_ttft` / `synthesize` before t0. |
+| HTTP TTS (ElevenLabs HTTP, OpenAI `tts-1-hd`) | TLS + TCP via a shared `httpx.AsyncClient` pre-warmed once per run (~80–200 ms). |
+
+HTTP-pool warming lives in `runner/src/coval_bench/providers/_http_session.py`.
+Providers opt in by overriding `Provider.warmup()` in `providers/base.py`; the
+orchestrator invokes warmup on every enabled provider class before the
+dataset loop runs and tears down the pool in the run's `finally` block.
+
+### HTTP/2 for the HTTP TTS cohort
+
+The shared pool uses HTTP/2 so that every concurrent request to a host
+multiplexes over a single connection. The connection is opened once by
+`warmup()`, and the dataset loop — run at a concurrency of 8 — reuses it, so
+TCP+TLS is paid once and excluded from every TTFA row rather than just the
+first. `http2=True` and the pool limits are set on `TimedTransport`, not on
+`AsyncClient`: httpx ignores both when a custom transport is supplied. Both
+hosts (`api.openai.com`, `api.elevenlabs.io`) negotiate HTTP/2; under 8-way
+concurrency the pool stays at one socket per host. Requires the `h2` package
+(`httpx[http2]`) — a build without it raises at client construction rather
+than downgrading silently.
+
+The protocol is chosen by TLS ALPN at connect time. It can negotiate HTTP/1.1
+instead — a TLS-terminating proxy or middlebox, an `HTTPS_PROXY` tunnel, or an
+HTTP/1.1-only endpoint. On HTTP/1.1 there is no multiplexing, so concurrent
+requests open additional connections that `warmup()` did not warm, and TTFA
+reabsorbs TCP+TLS for those rows. `warmup()` logs `http_version` and emits a
+`*_prewarm_no_http2` warning on any non-h2 connection, and every HTTP TTS
+result row records the negotiated `http_version`, so HTTP/1.1 rows can be
+filtered out of analysis rather than silently inflating the published TTFA.
+
+**Follow-ups, not yet in this implementation:**
+
+- *NVCF gRPC hosted cohort (Nvidia Nemotron, Magpie).* These providers ship a
+  per-channel warmup probe that excludes TLS + gRPC channel + NVCF GPU
+  dispatch (~500–1000 ms) before t0. The probe uses the same
+  `Provider.warmup()` hook documented above and will be wired in when
+  `nvidia_hosted.py` lands on `main`; the cohort table will gain a third row
+  then.
+- *Connect-time isolation at the transport layer.* `TimedTransport` records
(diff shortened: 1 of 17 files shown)


runner/tests/providers/tts/test_openai.py
```diff
@@ -214,6 +223,50 @@ def test_openai_name_property(fake_settings: Settings) -> None:
...
+    def handler(request: httpx.Request) -> httpx.Response:
+        seen.append(f"{request.method} {request.url.path}")
+        return httpx.Response(401, content=b"unauthorized")
+
+    _http_session._CLIENTS["openai"] = httpx.AsyncClient(
+        base_url="https://api.openai.com",
+        transport=httpx.MockTransport(handler),
+    )
+    await OpenAITTSProvider.warmup(fake_settings)
+    assert seen == ["HEAD /v1/models"]
```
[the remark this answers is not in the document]
seribaymadina: fixed the docstring to match implementation

runner/src/coval_bench/providers/_http_session.py
```diff
@@ -0,0 +1,104 @@
...
+    one-time warmup primes the pool, this interval should be small and
+    stable; large or growing values indicate the pool is reconnecting
+    mid-run.
+    """
+
+    async def handle_async_request(self, request: httpx.Request) -> httpx.Response:
+        t_submit = time.monotonic()
+        response = await super().handle_async_request(request)
+        t_headers = time.monotonic()
+        request.extensions["__t_submit"] = t_submit
```
lorenjphillips, line 61: these extensions are never read anywhere and the only output is the debug log below, so the guardrail isn't actually wired to anything. if the server drops the idle h2 connection mid-run a row silently reabsorbs connect time and nothing flags it. either surface submit-to-headers on the result row, or drop the stamps and stop calling it a guardrail

runner/src/coval_bench/providers/tts/openai.py
```diff
@@ -74,6 +80,26 @@ def name(self) -> str:
...
+        client = get_shared_client("openai", _BASE_URL)
+        t0 = time.monotonic()
+        response = await client.head("/v1/models")
+        logger.info(
+            "openai_prewarm",
+            warmup_ms=round((time.monotonic() - t0) * 1000, 1),
+            http_version=response.http_version,
+        )
+        if response.http_version != "HTTP/2":
+            logger.warning("openai_prewarm_no_http2", http_version=response.http_version)
```
lorenjphillips, line 104: on http/1.1 there's no multiplexing, so the 8 concurrent requests open cold connections warmup never touched and ttfa reabsorbs tcp+tls for those rows. a log warning isn't enough when the latency numbers are the deliverable. record http_version on each result so analysis can filter, or hard fail the run on no-h2. same applies to the elevenlabs warmup

runner/src/coval_bench/providers/tts/elevenlabs.py
```diff
@@ -54,64 +60,52 @@ def name(self) -> str:
...
-                output_format="pcm_24000",
-                text=text,
-                model_id=self._model,
-            )
-
-            for chunk in response:
-                if chunk:
-                    if self._is_audio_chunk(chunk) and ttfa_ms is None:
+            async with client.stream("POST", url, headers=headers, json=payload) as response:
+                response.raise_for_status()
```
lorenjphillips: raise_for_status on a streamed response raises before the body is read, so error rows lose elevenlabs' actual error message that the old sdk surfaced. read the body on the error path before raising

lorenjphillips (dismissed):
solid work and the methodology doc is genuinely good. three things to sort before merge, flagged inline - the warmth guardrail isn't actually wired to anything, the http/1.1 downgrade only warns when it invalidates the measurement, and streamed error rows lose the provider's error message

coval-cale (approved):
+1 to Loren's comments. Good catch to keep us from accidentally keeping silent failures in the metrics. But besides that this is so sick! Nice work!

runner/src/coval_bench/providers/_http_session.py
seribaymadina: wired it to record submit_to_headers_ms per row and flag new and reused connections and tested it

runner/src/coval_bench/providers/tts/elevenlabs.py
seribaymadina: kept the warning for early signal; http_version recorded per row can be filtered and http/1.1 rows hard fair and run goes PARTIAL hence gets dropped

runner/src/coval_bench/providers/tts/openai.py
seribaymadina: fix: read the body on is_error before returning so the error carries http {status}: {body}.  dropped the fix on row-level persistence as to avoid duplication with Cale's PR:  https://github.com/coval-ai/benchmarks/pull/47

runner/src/coval_bench/db/migrations/versions/20260601_0002_add_http_diagnostics.py
coval-cale, line 1: Good to check how we currently run migrations, is this handled by our infra or is there an extra manual step after this lands?

docs/methodology.md
```diff
@@ -100,6 +100,64 @@ The DP edit-distance computation is delegated to `jiwer`. Pinned in
...
+TTFT (STT) and TTFA (TTS) measure model inference time. Pre-t0 deployment
+setup — TCP, TLS, protocol handshake, optional session-setup RTT — is
+excluded for every provider. The rule is applied uniformly across cohorts;
+the magnitude of what each cohort excludes varies by deployment architecture,
+but the rule is the same.
+
+| Cohort | What is excluded from t0 |
+|---|---|
+| WS streaming (Deepgram, AssemblyAI, ElevenLabs Scribe, Speechmatics, Gradium STT, Cartesia, Deepgram aura-2, Rime, Gradium TTS, Hume) | TLS + WS upgrade + optional session-setup RTT (~50–200 ms). Handshake naturally completes inside `measure_ttft` / `synthesize` before t0. |
+| HTTP TTS (ElevenLabs HTTP, OpenAI `tts-1-hd`) | TLS + TCP via a shared `httpx.AsyncClient` pre-warmed once per run (~80–200 ms). |
```
coval-cale: We've depreciated OpenAI's tts-1-hd model. Switch to gpt-4o-mini-tts

coval-cale (approved):

runner/src/coval_bench/db/migrations/versions/20260601_0002_add_http_diagnostics.py
seribaymadina: valid concern, infra runs the migration at startup before benchmark run starts

docs/methodology.md
seribaymadina: thanks for catching it , fixed

PR #46, opened 2026-06-01
Fix dashboard model name collisions

seribaymadina:
Models that share a slug — Speechmatics and Gradium both expose `default` — were colliding in the dashboard, so one provider's series overwrote the other's. This keys charts and stats by `provider:model` so same-named models stay distinct, builds the sidebar list from actual results, and only plots models with data in the active window.

Commits:
- Composite `provider:model` keys across charts, stats, colors, and sidebar
- Normalize the provider name once in the chart axis tick (no behavior change)
- Sort heatmap rows by display name instead of the internal key

Known follow-ups (out of scope): latency-only models with no WER are dropped from WER-gated charts; catalogue models with no results can show as empty selections.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model name normalization and display consistency across all charts and tooltips.
  * Fixed model filtering logic to properly handle composite provider-model identifiers.
  * Corrected timeline data aggregation to accurately reflect selected models in current windows.

* **Refactor**
  * Simplified tooltip displays by removing unnecessary benchmark transcript data dependency.
  * Integrated provider information directly into tooltips for better model identification.
  * Updated performance metrics to use millisecond-based gap calculations instead of seconds.

* **Chores**
  * Removed Real-Time Factor (RTF) metric from performance comparisons.
  * Added support for Time-to-First-Token performance gap metric.

### DIFF
--- a/web/lib/utils/modelsFromResults.test.ts
+++ b/web/lib/utils/modelsFromResults.test.ts
@@ -1,0 +1,129 @@
+// Copyright 2026 The Coval Benchmarks Authors
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it } from "vitest";
+import type { Result } from "../aggregates";
+import { buildModelsByProviderFromResults } from "./modelsFromResults";
+
+function row(
+  provider: string,
+  model: string,
+  benchmark: "STT" | "TTS" = "STT"
+): Result {
+  return {
+    id: 1,
+    provider,
+    model,
+    benchmark,
+    metric_type: "TTFT",
+    metric_value: 0.5,
+    metric_units: "seconds",
+    status: "SUCCEEDED",
+    created_at: "2026-05-18T12:00:00Z",
+    run_id: 1,
+    audio_filename: "a.wav",
+    transcript: null,
+    error: null,
+    voice: null,
+  } as Result;
+}
+
+describe("buildModelsByProviderFromResults", () => {
+  it("includes providers/models present in result rows when no catalogue is available", () => {
+    const out = buildModelsByProviderFromResults(
+      [
+        row("deepgram", "nova-2"),
+        row("xai", "grok-stt"),
+      ],
+      "STT"
+    );
+    expect(out.deepgram).toEqual(["deepgram:nova-2"]);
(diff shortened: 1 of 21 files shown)


web/next-env.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/types/routes.d.ts";
+import "./.next/dev/types/routes.d.ts";
```
lorenjphillips: This file is generated and should not be touched in this PR. `next build` rewrites this import back to `./.next/types/routes.d.ts`, so this branch will keep creating local churn for anyone who builds it. Drop `web/next-env.d.ts` from the PR.

web/lib/utils/modelsFromResults.test.ts
```diff
@@ -0,0 +1,153 @@
...
+    const out = buildModelsByProviderFromResults(
+      [
+        row("deepgram", "nova-2"),
+        row("xai", "grok-stt"),
+      ],
+      "STT"
+    );
+    expect(out.deepgram).toEqual(["deepgram:nova-2"]);
+    expect(out.xai).toEqual(["xai:grok-stt"]);
+    expect(out.openai).toBeUndefined();
```
lorenjphillips, line 42: The duplicate-slug case this PR fixes needs a unit test here. Add two result rows with different providers and the same `model` slug, e.g. `speechmatics/default` and `gradium/default`, and assert the output contains both `speechmatics:default` and `gradium:default`. Current tests cover result-backed models and disabled filtering, but not the collision path that caused the dashboard overwrite.

web/next-env.d.ts
seribaymadina: dropped it

web/lib/utils/modelsFromResults.test.ts
seribaymadina: added two rows to assert speechmatics default and gradium default to come through as distinct

web/next-env.d.ts
coval-cale: This one gets me all the time

coval-cale (approved):
Yewww!

PR #47, opened 2026-06-01
[BENCH-219] Fix silent errors.

coval-cale:
- Errors weren't being properly handled.
- Providers caught errors and passed them to the result object.
- Orchestrator expected errors to be propagated.
- This is a quick fix that checks the result object for an error instead of expecting that error to be propagated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Metrics now fail with clear messages when provider errors or missing values occur; downstream rows (including timing and RTF/TTFA) reflect those failures and affect run status (FAILED/PARTIAL/SUCCESS).
  * WER scoring is emitted only when required upstream data exists (complete transcript or synthesized audio); WER rows are suppressed otherwise.

* **Tests**
  * Added regression tests covering provider exceptions, empty/partial/errored STT/TTS results, WER failures, and related run-status behavior.

### DIFF
--- a/runner/src/coval_bench/runner/orchestrator.py
+++ b/runner/src/coval_bench/runner/orchestrator.py
@@ -85,6 +85,30 @@ def _truncate(msg: str) -> str:
     return msg[:_MAX_ERROR_LEN]
 
 
+def _metric_outcome(
+    metric_value: float | None,
+    item_error: str | None,
+    metric_label: str,
+    result_status: Any,  # noqa: ANN401 — ResultStatus enum, lazy-imported by callers
+) -> tuple[Any, str | None]:
+    """Decide ``(status, error)`` for a single result row.
+
+    A provider that returns without raising can still have failed: it may set
+    ``result.error`` or simply produce no measurement. The orchestrator used to
+    stamp every returned row ``SUCCESS``; this restores honest labelling.
+
+    Rules (in order):
+    * an item-level provider error fails the row and carries the message;
+    * otherwise a missing (``None``) metric fails just this row;
+    * a present metric succeeds.
+    """
+    if item_error:
+        return result_status.FAILED, _truncate(item_error)
+    if metric_value is None:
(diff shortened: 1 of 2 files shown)


runner/src/coval_bench/runner/orchestrator.py
```diff
@@ -207,71 +233,76 @@ async def _run_stt_item(
...
-                    transcript=None,
-                    status=ResultStatus.FAILED,
-                    error=err,
-                )
-            )
-            return results
+
+        # Single failure model: a raised exception OR a provider-reported error fails every
+        # metric row below (see _metric_outcome). Read measurements defensively — the call
+        # may have raised before producing a result.
```
lorenjphillips: this seems self documenting can we drop the comment?

runner/tests/unit/test_orchestrator.py
```diff
@@ -952,3 +958,193 @@ async def _send_sigterm_after_start() -> None:
     finish_kwargs = writer.finish_run.await_args.kwargs
     assert finish_kwargs["status"] == RunStatus.PARTIAL
     assert "sigterm" in (finish_kwargs.get("error") or "").lower()
+
+
+# ---------------------------------------------------------------------------
+# 14-17. BENCH-219 — degraded results that return (no raise) are marked FAILED
```
lorenjphillips: Can we remove linear ticket references from OSS repo?

lorenjphillips (approved):
approved but some comments to check out

lorenjphillips (approved):
Looks good.

runner/src/coval_bench/runner/orchestrator.py
seribaymadina, line 341: to keep in mind: WER failure emits no row so a compute_wer crash seems to  leave the run succeeded with WER absent. there is no user facing impact but the run status signal is lost we could consider a failed WER row.

seribaymadina:
amazing job done, thanks for the fix!

seribaymadina (approved):

PR #48, opened 2026-06-02
Fix deepgram wer

coval-cale:
## Summary by CodeRabbit

* **Bug Fixes**
  * Switched Deepgram websocket usage to the v2 listen endpoint; removed incompatible query parameters and added a runtime check that Flux models require mono audio.
  * Revised streaming transcript assembly to store and order per-turn finals correctly and to preserve trailing interim text when the stream ends.

* **Tests**
  * Added tests covering Flux URL validation (mono), multi-turn final concatenation, is_final handling, and inclusion of unfinalized trailing transcripts.

### DIFF
--- a/runner/src/coval_bench/providers/stt/deepgram.py
+++ b/runner/src/coval_bench/providers/stt/deepgram.py
@@ -163,6 +163,7 @@ class DeepgramProvider(STTProvider):
 
     async def _receive(self, ws: Any, result: TranscriptionResult) -> None:
         final_segments: list[str] = []
+        pending_partial: str = ""
         flux_latest: str = ""
         last_final_time: float | None = None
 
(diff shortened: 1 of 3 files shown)


runner/src/coval_bench/providers/stt/deepgram.py
```diff
@@ -248,11 +249,13 @@ async def _receive(self, ws: Any, result: TranscriptionResult) -> None:
 
         # Build complete transcript
         if self._model in ("flux-general-en", "flux-general-multi"):
-            result.complete_transcript = flux_latest.strip() or None
-        elif final_segments:
-            result.complete_transcript = " ".join(final_segments).strip()
-        elif result.partial_transcripts:
-            result.complete_transcript = max(result.partial_transcripts, key=len).strip() or None
+            joined = " ".join(flux_turns[i] for i in sorted(flux_turns)).strip()
```
seribaymadina, line 252: This now stitches every Flux turn together instead of keeping only the last one, but the single Flux fixture uses one turn — so if anyone reverts to last-wins, the suite stays green and the regression slips through. If the BENCH-220 numbers show Flux WER actually dropped from this change then multi-turn is real and a quick two-turn test would lock it in; if Flux was flat, this is just future-proofing and can wait.

seribaymadina, line 252: joins all flux turns together instead of just keeping the last, but the only flux test uses one turn, so a revert to the old behavior wouldn't fail the tests. worth a quick two turn test

seribaymadina (approved):
thanks! one non-blocking comment

runner/src/coval_bench/providers/stt/deepgram.py
coval-cale: Done! Thanks!

PR #25, opened 2026-05-20
feat(providers/stt): add xAI Grok STT

seribaymadina:
Adds the xAI Grok streaming STT provider (`grok-stt`) to the runner.

- WebSocket provider with realtime audio pacing.
- Force-finalizes on `audio.done`, then waits for the terminal `transcript.done` before closing the socket (5s fallback) so the final can't be lost to the close. Verified live that Grok emits the terminal at end of stream regardless of endpointing, so `endpointing=200` stays for the first-partial number.
- Registry, chart color, and model/provider label entries for Grok STT.
- Tests and fixtures covering success, empty done, error, and the close gate.

Branch is synced with current main, so the diff is just the Grok additions.

<h3>Greptile Summary</h3>

This PR adds the xAI Grok STT WebSocket provider (`grok-stt`) to the benchmarks runner, following the same patterns established by other streaming providers (Deepgram, AssemblyAI, etc.). It also removes the unused `audio_duration` parameter from the `STTProvider.measure_ttft` base signature across all providers.

- **New `XaiSTTProvider`** streams 16-bit PCM audio over WebSocket to `wss://api.x.ai/v1/stt`, handles `transcript.partial`/`transcript.done`/`error` events, and implements a close gate that waits up to 5 seconds for `transcript.done` before closing the socket, preventing the terminal event from being lost.
- **Registry, chart color, and model/provider label** entries are added to `config.py`, `colors.ts`, `models.ts`, and `formatters.ts`, consistent with how other providers are registered.
- **Test suite** covers success, empty-done fallback, missing-terminal error, error-event propagation, URL/header setup, and deadline behaviour via `FakeWebSocket` fixture mocking.

<h3>Confidence Score: 5/5</h3>

Safe to merge — the new provider is well-isolated and the close gate, error-preservation, and transcript-finalization paths all behave correctly.

The xAI provider is a clean addition with no broken contracts. The close gate correctly waits for transcript.done before closing the socket. result.error = result.error or str(exc) in measure_ttft preserves the server's actionable error message over secondary WebSocket closure noise. The base-class signature removal of audio_duration is propagated consistently across all call sites and implementations. The one observation (extra sleep after the last audio chunk) is a minor metric-inflation concern, not a correctness issue.

No files require special attention; xai.py and its test suite are the core of the change and look correct.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
[...]

PR #49, opened 2026-06-02
Add OpenAI gpt-realtime-whisper STT provider

seribaymadina:
Adds the OpenAI Realtime API STT provider (gpt-realtime-whisper) over WebSocket, wired into the registry, run matrix, and web display config.

Notes:
- `turn_detection` is `null` — the model rejects `server_vad`, so it's single-pass manual-commit. TTFT stays comparable to native streaming; audio-to-final carries a post-audio commit round-trip.
- A `failed`/`error` event cancels the sender immediately instead of waiting for the full clip to stream.
- scipy's resampler is warmed at import so its lazy init never lands in a measured run.
- `_transcript_utils.py` is a new helper used only by this provider; PR #25 (Grok STT) adds an identical copy — coordinate merge order.

## Release Notes

* **New Features**
  * Added OpenAI's GPT Realtime Whisper as a new speech-to-text provider, enabling real-time audio transcription via websocket streaming.

* **Chores**
  * Added scipy as a dependency for audio processing.
  * Updated UI configuration to display the new OpenAI model with proper branding and colors.
  * Added comprehensive test coverage for the new provider.


web/next-env.d.ts
coval-cale, line 1: Did you mean to delete?

runner/src/coval_bench/providers/stt/openai.py
```diff
@@ -0,0 +1,266 @@
...
+                    if not transcript:
+                        continue
+                    set_first_token(result, transcript, now=now)
+                    add_partial_transcript(result, transcript)
+                    completed_transcript = transcript
+                    if result.audio_start_time is not None:
+                        last_final_time = now
+                    break
+
+                if event_type == "conversation.item.input_audio_transcription.failed":
```
coval-cale, line 248: Error here wouldn't properly propagate up to measure_ttft so it could end up waiting for a while.

runner/src/coval_bench/providers/stt/openai.py
```diff
@@ -0,0 +1,266 @@
...
+import math
+import time
+from collections import defaultdict
+from typing import Any, cast
+
+import numpy as np
+import structlog
+import websockets.asyncio.client as ws_client
+from pydantic import SecretStr
+from scipy import signal
```
coval-cale, line 29: Make sure this doesn't load lazy or we'll get the revenge of the wavy graphs :)

runner/src/coval_bench/providers/stt/_transcript_utils.py
coval-cale, line 1: I think this is also introduced in the grok PR. Might want to go back over the grok before you merge as there will probably be some conflicts.

coval-cale (approved):

runner/src/coval_bench/providers/stt/_transcript_utils.py
seribaymadina: yes, I want to get it in via this PR!

runner/src/coval_bench/providers/stt/openai.py
seribaymadina: thanks for the catch : added warmup at import as an insurance; it shouldn't affect TTFT/ audio to final though!

web/next-env.d.ts
seribaymadina: yep

PR #51, opened 2026-06-03
Fix/Warm librosa before traffic to fix inflated TTFA

coval-cale:
## Summary by CodeRabbit

* **Improvements**
  * Ensure audio-related libraries are loaded before TTS benchmark provider warmup when TTS is enabled and providers exist, improving reliability; failures during this warmup are non-fatal.

* **Tests**
  * Added tests verifying warmup runs before TTS provider traffic, is skipped for STT-only runs, includes an integration-style real-call, and tolerates warmup import failures without failing the run.

### DIFF
--- a/runner/src/coval_bench/runner/orchestrator.py
+++ b/runner/src/coval_bench/runner/orchestrator.py
@@ -169,6 +169,20 @@ def _get_metrics() -> tuple[Any, Any]:
     return mod.compute_wer, mod.compute_rtf
 
 
+def _warm_audio_metrics() -> None:
+    """Absorb librosa's lazy numba/scipy import before any provider traffic.
+
+    The first ``first_audible_offset_ms`` call pulls in the stack (~18s on one
+    Cloud Run vCPU); mid-phase it freezes the event loop and inflates every
+    in-flight TTFA. Must run before provider warmup so the freeze can't idle
+    out freshly warmed connections, and before the run row exists — the loop
+    can't service SIGTERM during the freeze, so a row would strand at
+    'running'.
+    """
+    mod = importlib.import_module("coval_bench.metrics")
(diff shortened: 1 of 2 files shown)


runner/src/coval_bench/runner/orchestrator.py
```diff
@@ -696,6 +708,9 @@ def _on_sigterm() -> None:
             # open, NVCF dispatch).  Default is a no-op; only providers that
             # need it pay the call.  Errors are non-fatal.
             # ------------------------------------------------------------------
+            if benchmark_kind in ("tts", "both") and enabled_tts:
+                _warm_audio_metrics()
```
seribaymadina: the entire runtime cost of librosa traces to one helper call:  slicing PCM into overlapping windows. And librosa.util.frame is trivially replaceable with numpy's sliding_window_view. That would make the whole runtime librosa-free and eliminate the need for _warm_audio_metrics() altogether better fix for long-term.

runner/src/coval_bench/runner/orchestrator.py
```diff
@@ -169,6 +169,18 @@ def _get_metrics() -> tuple[Any, Any]:
     return mod.compute_wer, mod.compute_rtf
 
 
+def _warm_audio_metrics() -> None:
+    """Absorb librosa's lazy numba/scipy import before any provider traffic.
+
+    The first ``first_audible_offset_ms`` call pulls in the stack (~18s on one
+    Cloud Run vCPU); mid-phase it freezes the event loop and inflates every
+    in-flight TTFA. Must run before provider warmup so the freeze can't idle
```
seribaymadina, line 177: note: only openai tts and elevenlabs have real warmup, others don't

seribaymadina (approved):
left non blocking comments for short-term solution; for better fix, we might consider runtime to be "librosa-free"

PR #54, opened 2026-06-03
No more librosa

coval-cale:
Drop librosa. Was used in audio.py but nothing else used audio.py so dropped because dead code.

## Summary by CodeRabbit

* **Refactor**
  * Refactored audio processing implementation to use optimized NumPy-based algorithms.
  * Streamlined audio metric calculations for improved performance.
  * Updated test cases to reflect implementation changes.


seribaymadina (approved):
lgtm!

PR #55, opened 2026-06-03
Stop filter-menu toggles from re-rendering all charts

seribaymadina:
The filter menu was laggy because everything shared one context — expanding a provider or opening the mobile sheet re-rendered every chart, even though the charts don't use menu state.

Two fixes:
- Menu state (expanded providers, mobile sheet) now lives in its own context, so menu toggles only re-render the sidebar.
- Chart aggregations are memoized, so they recompute only when the data or selection actually changes.

No visual change — it's a performance fix.


coval-cale (approved):
yew!

robyoungsf (approved):
Love it thank you!

PR #62, opened 2026-06-04
Add PostHog analytics to dashboards and playground

seribaymadina:
Adds PostHog to the public benchmark site so we can see which models and metrics people actually engage with.

- Tags every event `product: "public-benchmarks"`; autocapture and session recording are off, so only named events fire.
- Wires dashboard + playground interactions (model selection, chart hover, scroll, sort, benchmark runs, mode switch, etc.). Page visits come free via `$pageview`.
- Safe props only — counts, ids, durations — never raw text or audio.

Needs `NEXT_PUBLIC_POSTHOG_TOKEN` (public-benchmarks key, already in Vercel).

## Summary by CodeRabbit

* **Chores**
  * Integrated analytics infrastructure to capture user interaction and engagement metrics across dashboard and playground features, including chart interactions, scrolling patterns, and mode selections. Optional analytics can be disabled by leaving configuration unset.


web/lib/posthog/client.ts
coval-cale, line 1: Add the license header to all the new files. Not sure if this is important but might as well match convention.

```
// Copyright 2026 The Coval Benchmarks Authors
// SPDX-License-Identifier: Apache-2.0
```

web/lib/posthog/events.ts
```diff
@@ -0,0 +1,32 @@
...
+  dashboardScrollDepth: "dashboard_scroll_depth",
+  playgroundTtsTypingStarted: "playground_tts_typing_started",
+  dashboardHeatmapSorted: "dashboard_heatmap_sorted",
+  dashboardWerBarClicked: "dashboard_wer_bar_clicked",
+  dashboardChartPanned: "dashboard_chart_panned",
+  sttTranscriptBrowsed: "stt_transcript_browsed",
+  playgroundExamplePromptUsed: "playground_example_prompt_used",
+  playgroundModeSwitched: "playground_mode_switched",
+  playgroundResultPlayed: "playground_result_played",
+  resultShared: "result_shared"
```
coval-cale: Is this used anywhere? If not let's drop it.

web/app/playground/components/playground-draft/STTPlaygroundPanel.tsx
```diff
@@ -174,6 +197,14 @@ export function STTPlaygroundPanel({
   const scrollCarouselStep = useCallback((dir: -1 | 1) => {
     const vp = carouselViewportRef.current;
     if (!vp) return;
+    if (!transcriptBrowsedFiredRef.current) {
+      transcriptBrowsedFiredRef.current = true;
+      capturePostHogEvent(POSTHOG_EVENTS.sttTranscriptBrowsed, {
+        surface: "playground",
+        mode: "stt",
+        method: "arrow"
```
coval-cale, line 205: I think this only captures scrolling done with the arrow keys and won't capture stuff like mouse-wheel scrolling.

web/app/playground/hooks/usePlaygroundMode.ts
```diff
@@ -43,6 +45,12 @@ export function usePlaygroundMode(): {
     (next: PlaygroundMode) => {
       if (next === mode) return;
 
+      capturePostHogEvent(POSTHOG_EVENTS.playgroundModeSwitched, {
+        surface: "playground",
+        from: mode,
+        to: next
```
coval-cale: No trigger here so posthog won't catch.

web/components/layout/DashboardLayout.tsx
```diff
@@ -3,15 +3,44 @@
...
 
 const DashboardLayout: React.FC<{ children: React.ReactNode }> = ({ children }) => {
   const { sidebarCollapsed, loading } = useDashboard();
+  const mode = useActiveTab();
+  const firedDepthsRef = useRef<Set<number>>(new Set());
+
+  useEffect(() => {
+    firedDepthsRef.current.clear();
+    const thresholds = [25, 50, 75, 100];
+    const onScroll = () => {
```
coval-cale, line 24: I think adding this check immediately has the potential to double count. For example if a user scrolls down to the bottom of the page, it'll trigger at thresholds [25, 50, 75, 100]. Awesome! But, and I might be wrong, if a user then reloads the page, this will get called on reload, the page will autoscroll them to the bottom and this listener will fire again on all four thresholds.

Honestly, I'm not super worried about this. Maybe we just flag this in linear and go back to it if our scroll metrics look weird.

coval-cale (approved):
Looks good! Added a couple nits, nothing critical. Your call on whether to address or not.

web/lib/posthog/client.ts
seribaymadina: keeping it for convention

web/lib/posthog/events.ts
seribaymadina: good call!

seribaymadina: dropped

web/components/layout/DashboardLayout.tsx
seribaymadina: dropped the on mount onScroll() call so a reload that restores this to the bottom no longer retires all the threshold contain real scrolls, thanks for the feedback

PR #64, opened 2026-06-04
Add PostHog analytics to the API and runner

seribaymadina:
## Summary

Server-side PostHog analytics, the backend half of #62 (which is browser-only).

- API routers capture one event per query: leaderboard, providers, results, runs.
- Orchestrator captures run completed, failed, and SIGTERM events.
- Gated on `POSTHOG_PROJECT_TOKEN`; off when unset.

## Testing

`unittest.mock` + `dependency_overrides`, no new deps, no network:

- Lifespan builds the client, the real `get_posthog` returns it, shutdown flushes.
- Every API event name and payload, including the `$process_person_profile` guard.
- Orchestrator completed and failed events.
- Tests disable PostHog by default, so they never use a local `.env` token.

SIGTERM event untested by design (needs a flaky mid-run `CancelledError`).

Gate green: ruff, ruff format --check, mypy --strict src.

## Summary by CodeRabbit

* **New Features**
  * Added analytics tracking for API endpoints and benchmark runs; events emitted for provider lists, runs, results queries, and leaderboard access.

* **Configuration**
  * New analytics settings: project token, host URL, and disable toggle (configurable via environment variables).

* **Tests**
  * Added end-to-end and unit tests covering analytics initialization, per-route event emission, and tolerant failure behavior.

* **Chores**
  * .gitignore updated to exclude local environment variants (.env*.local).

### DIFF
--- a/runner/pyproject.toml
+++ b/runner/pyproject.toml
@@ -38,6 +38,8 @@ dependencies = [
     # WER
     "jiwer>=3.0",
     "whisper-normalizer>=0.0.10",
+    # Analytics
+    "posthog>=3.0",
 ]
 
 [project.optional-dependencies]
@@ -126,6 +128,10 @@ ignore_missing_imports = true
 
 [[tool.mypy.overrides]]
 module = "aiohttp.*"
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "posthog.*"
 ignore_missing_imports = true
 
 [tool.coverage.run]
--- a/runner/src/coval_bench/api/routers/leaderboard.py
+++ b/runner/src/coval_bench/api/routers/leaderboard.py
@@ -26,7 +26,7 @@ from posthog import Posthog
 from psycopg_pool import AsyncConnectionPool
 from starlette.requests import Request
 
-from coval_bench.api.deps import get_pool, get_posthog
+from coval_bench.api.deps import capture_api_event, get_pool, get_posthog
 from coval_bench.api.ratelimit import limiter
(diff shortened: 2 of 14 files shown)


runner/src/coval_bench/api/routers/leaderboard.py
```diff
@@ -124,4 +126,16 @@ async def get_leaderboard(
         entry_rows = await rows.fetchall()
 
     entries = [LeaderboardEntry.model_validate(r) for r in entry_rows]
+    if posthog_client is not None:
+        posthog_client.capture(
```
coval-cale: I could be wrong but I think this will raise an error. I'm pretty sure only the event is positional so the "coval-bench-api" will be unexpected.

runner/pyproject.toml
```diff
@@ -38,6 +38,8 @@ dependencies = [
     # WER
     "jiwer>=3.0",
     "whisper-normalizer>=0.0.10",
+    # Analytics
+    "posthog>=3.0",
```
coval-cale, line 42: Posthog is currently on V7, maybe bump this up to >=v6.

coval-cale (changes requested):
Check the capture() function. Good to go after that.

runner/src/coval_bench/api/routers/leaderboard.py
coval-cale: Looks like this is in a couple places.

coval-cale (approved):
lgtm!

PR #66, opened 2026-06-04
Aggregate latency vs wer datapoints

coval-cale:
## Summary by CodeRabbit

* **UI/UX Improvements**
  * Reordered dashboard sections on STT and TTS pages for clearer priority.

* **Tooltips**
  * Labels clarified with "Avg" prefixes, multi-line layout, and added an explicit "Runs" count.

* **Data Visualization**
  * Charts now show run-weighted averages per model, adjust axis scales/ticks to actual data, use consistent point grouping, and display all measurements rather than summarizing outliers.

### DIFF
--- a/web/hooks/useChartData.ts
+++ b/web/hooks/useChartData.ts
@@ -299,7 +299,8 @@ export function useChartData({
     [violinDataMemo]
   );
 
-  // Scatter needs paired per-run values — cannot use pre-aggregated stats
+  // Pair latency with WER per run, then average per model — pairing restricts
+  // the average to runs that produced both metrics
   const scatterDataMemo = useMemo<ScatterDataPoint[]>(() => {
     const selectedModels =
       activeTab === "tts" ? selectedTTSModels : selectedSTTModels;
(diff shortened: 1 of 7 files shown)


web/hooks/useChartData.ts
```diff
@@ -319,8 +319,10 @@ export function useChartData({
     rawData.forEach((item) => {
       if (!selectedModelKeys.has(toModelKey(item.provider, item.model))) return;
       if (item.metric_type !== xMetric && item.metric_type !== yMetric) return;
+      if (!INCLUDED_STATUSES.has(item.status)) return;
+      if (item.metric_value === null || item.metric_value === undefined) return;
 
-      const key = `${item.benchmark}_${item.provider}_${item.model}_${item.timestamp}`;
+      const key = `${item.benchmark}_${item.provider}_${item.model}_${item.voice}_${item.audio_filename}_${item.scheduled_at}`;
```
seribaymadina: non blocking : seems like could be paired on run_id because otherwise in some cases first pair would be the only one kept;

seribaymadina (approved):
left one non blocking comment

PR #68, opened 2026-06-04
Get rid of chart scrolling cause it didn't work

coval-cale:
Apparently we could scroll through our charts (theoretically) but not in real life so I dropped that to simplify things.

## Summary by CodeRabbit

* **Refactor**
  * Timeline window now derives automatically from the latest data (no external anchoring).
  * Chart interactions simplified: panning/drag-to-scroll removed; charts use hover for tooltips and have a fixed display height.
  * Dashboard state no longer exposes dragging/anchor controls, improving consistency across views.


seribaymadina (approved):
thanks for catching this

PR #69, opened 2026-06-05
Make model select/deselect responsive

seribaymadina:
Toggling a model re-computed every selected model's chart data on every click, so it got slower as data grew. Now each model's series is built once on load, and a toggle only touches the clicked model — the cost stays flat as we add providers, models, and history.

- Per-model precompute for scatter/timeline/violin/heatmap
- Key-metrics memoized + grouped in one pass (was re-filtering per model every render)
- Charts read a deferred selection — checkbox responds instantly
- Chart animations off; quickselect instead of full-sort median

Same numbers shown, just computed cheaply. Page-load work is a separate follow-up.

## Summary by CodeRabbit

* **Performance**
  * Refactored chart data pipeline to precompute model aggregates, improving responsiveness when selecting and switching between models and analysis tabs.
  * Deferred model selection updates to reduce UI rendering overhead during rapid interactions.
  * Disabled animations on bar, scatter, and timeline charts for faster visual updates.

### DIFF
--- a/web/components/dashboard/AccuracyBarSection.tsx
+++ b/web/components/dashboard/AccuracyBarSection.tsx
@@ -112,6 +112,7 @@ const AccuracyBarSection: React.FC = () => {
                 stroke={themeColors.barStroke}
                 strokeWidth={1}
                 radius={[4, 4, 0, 0]}
+                isAnimationActive={false}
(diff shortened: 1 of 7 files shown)


coval-cale (approved):
Couldn't even find a nit! lgtm!

PR #73, opened 2026-06-05
Add results created_at index

coval-cale:
Our db is getting pretty large and we weren't indexing on the main query for STT/TTS pages. Led to some slow down. This fixes a bit of that slowdown.

## Summary by CodeRabbit

* **Documentation**
  * Corrected internal routing notes to reference the proper index used for 24‑hour window result queries.

* **Infrastructure**
  * Added a non-blocking database index to improve performance of benchmark result and leaderboard queries filtered by creation time.

### DIFF
--- a/runner/src/coval_bench/db/migrations/versions/20260605_0004_add_results_benchmark_created_at_idx.py
+++ b/runner/src/coval_bench/db/migrations/versions/20260605_0004_add_results_benchmark_created_at_idx.py
@@ -1,0 +1,33 @@
+# Copyright 2026 The Coval Benchmarks Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Add (benchmark, created_at) index on results.
+
+The TTS/STT dashboards (GET /v1/results) and the live 7d/30d leaderboard
+filter results by benchmark plus a created_at window. No existing index
+covers that path — results_provider_model_idx leads with provider/model,
+which those queries don't filter on — so they seq-scan the whole table.
+"""
+
+from __future__ import annotations
+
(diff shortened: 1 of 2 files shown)


runner/src/coval_bench/db/migrations/versions/20260605_0004_add_results_benchmark_created_at_idx.py
```diff
@@ -0,0 +1,33 @@
...
+down_revision = "20260602_0003"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """Add the (benchmark, created_at DESC) index."""
+    op.execute("SET search_path TO benchmarks_v2")
+    op.execute(
+        "CREATE INDEX results_benchmark_created_at_idx ON results (benchmark, created_at DESC)"
```
seribaymadina: might consider : ( low-risk one time, so decision is up to you) CREATE INDEX CONCURRENTLY inside an Alembic autocommit_block since locks out writes for the whole build. Since migrations run at Cloud Run Job boot and the table is getting pretty large, that lock duration grows over tisome and can block/delay the runner's writes.

seribaymadina (approved):
clean! left one comment that you might want to consider making a small decision on

runner/src/coval_bench/db/migrations/versions/20260605_0004_add_results_benchmark_created_at_idx.py
coval-cale: nice catch! Will do!

PR #80, opened 2026-06-08
Fix/overview uses aggregates

seribaymadina:
Overview leaderboards now read the live `/v1/results/aggregates` endpoint instead of the stale `results_24h` materialized view. Also: `/` lands on `/overview`, not`/tts`.

Same endpoint, params, and ranking fields as the pages (TTS by p50 TTFA, STT by avg WER), so they can't drift apart.

Note:
Removes the last caller of `/v1/leaderboard` (it only powered these cards). The endpoint + `results_24h` MV are now orphaned. Deleting them is out of scope for this PR.

## Summary by CodeRabbit

* **New Features**
  * Leaderboards completely redesigned to display top-performing models with metrics specifically optimized for TTS and STT use cases. The new interface provides enhanced visibility into model rankings, improved performance data presentation, and streamlined model comparison capabilities to help users make better-informed selection decisions
  * Home page navigation enhanced with improved routing and user experience optimization

### DIFF
--- a/web/app/page.tsx
+++ b/web/app/page.tsx
@@ -4,5 +4,5 @@
 import { redirect } from "next/navigation";
 
 export default function Home() {
-  redirect("/tts");
+  redirect("/overview");
 }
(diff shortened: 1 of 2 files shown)


coval-cale (approved):
Nice work!

PR #82, opened 2026-06-08
Add ttl cache

coval-cale:
- Adds a TTL cache to the aggregation query. Simple fix to speed up load times. 
- Cache expires after 15 mins so we're never that far behind live.


seribaymadina (approved):
great!

PR #85, opened 2026-06-09
Add TTFS streaming-STT metric

seribaymadina:
Adds TTFS — time from end-of-speech to the final transcript, anchored at a shared VAD offset pinned per clip. Measuring from speech end keeps latency comparable across providers and clip lengths.

- Runner trims trailing silence and emits a TTFS row (a missing offset or failed calc fails the row, not a silent null)
- Served by the results + leaderboard APIs
- STT timeline gets a TTFS/TTFT toggle, default TTFS

Follow-ups (separate PRs): Flux/AssemblyAI endpointing parity, and a TTFS toggle on the latency-vs-accuracy chart.

## Release Notes

* **New Features**
  * Added TTFS (Time to Final Segment) metric to the leaderboard, accessible alongside existing latency metrics.
  * Introduced metric selector in timeline chart allowing users to toggle between TTFS and TTFT for STT benchmarks.
  * Added VAD offset preprocessing capability for audio dataset preparation.

* **Updates**
  * Extended leaderboard and results API endpoints to support the new TTFS metric option.
  * Updated STT manifest schema to include optional speech end-offset timestamps.

<h3>Greptile Summary</h3>

Introduces TTFS (Time to Final Segment) as a new STT latency metric anchored at a shared per-clip VAD end-of-speech offset, enabling provider comparisons that are independent of clip length and speaker pace.

- Runner trims trailing silence from each clip using the precomputed `speech_end_offset_ms` before dispatch, then emits a TTFS result row; a missing offset or negative result fails the row (never a silent null).
- Backend: `compute_ttfs` in `metrics/ttft.py`, manifest schema extended, leaderboard and results APIs updated with TTFS as a valid metric.
- Frontend: `useChartData` refactored to index timeline series per-metric; `TimelineChart` gains a TTFS/TTFT toggle for the STT tab, defaulting to TTFS.

<h3>Confidence Score: 4/5</h3>

Safe to merge after fixing the tab-switch metric state in TimelineChart — the runner, metric, and API changes are correct.

The backend additions (trimming, TTFS computation, manifest, APIs) are clean and well-tested. The frontend refactor in useChartData is solid. The one concrete defect is in TimelineChart: metric state is initialized from activeTab only once at mount, so switching between STT and TTS leaves the chart querying the wrong metric — blank chart for TTS if the user navigated from STT, and an un-highlighted toggle for STT if navigating from TTS with no way to recover without a page reload.

[...]

PR #88, opened 2026-06-09
Add TTFS/TTFT toggle to latency-vs-accuracy chart

seribaymadina:
Adds a TTFS/TTFT toggle to the latency-vs-accuracy scatter, matching the timeline toggle. Also defaults both STT latency charts (timeline + scatter) to **TTFT for now** — TTFS has no data until #85 deploys and runs accumulate, so TTFT avoids an empty chart on load. Flip the defaults back to `"TTFS"` in ~a day (one-liner each, marked by a comment).

- `getScatterData` is parameterized by metric (one-pass `scatterByMetricModel` + thin getter, same shape as the timeline)
- The chart derives its x-axis and labels from the selected metric; STT-only, TTS stays TTFA
- No Pareto frontier — matches the current chart
- 2nd commit defaults the **timeline** toggle to TTFT too (it lives on `main` from #85, so it rides here rather than a separate PR)

Builds on the TTFS metric (#85, merged).

### DIFF
--- a/web/hooks/useChartData.ts
+++ b/web/hooks/useChartData.ts
@@ -239,19 +239,26 @@ export function useChartData({
     [boxPlotDataMemo]
   );
 
-  // One (avg latency, avg WER) point per model. The averages are independent
-  // per metric — not restricted to runs that produced both.
-  const scatterByModel = useMemo<Record<string, ScatterDataPoint>>(() => {
-    const xMetric = activeTab === "tts" ? "TTFA" : "TTFT";
-
-    const byModel: Record<string, ScatterDataPoint> = {};
+  // Per-metric, per-model scatter point (avg latency, avg WER) built in one
+  // pass, indexed by metric_type so any latency metric can read its own points.
+  // Averages are independent per metric — not restricted to runs producing both.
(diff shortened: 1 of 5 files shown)


web/hooks/useChartData.ts
```diff
@@ -264,26 +263,19 @@ export function useChartData({
...
-        points.push(point);
-      }
-    });
-    return points;
-  }, [scatterByModel, activeTab, selectedTTSModels, selectedSTTModels]);
-
   const getScatterData = useCallback(
-    (): ScatterDataPoint[] => scatterDataMemo,
-    [scatterDataMemo]
+    (metric: string): ScatterDataPoint[] => {
```
coval-cale, line 278: Repeat of the first toggle. Might as well abstract it out into a shared class? Idk I don't feel super strongly about it.

coval-cale (approved):
YEWWWW!

web/hooks/useChartData.ts
seribaymadina: I'm guessing we will do a lot of refactoring for it, so leaving it for now

PR #90, opened 2026-06-10
Force STT finalization at end-of-speech for TTFS parity

seribaymadina:
AssemblyAI sends ForceEndpoint before Terminate (with a high end_of_turn_confidence_threshold), and nova disables native endpointing and sends Finalize before CloseStream, so both finalize on the shared VAD end-of-speech signal rather than their own turn decision. Flux can't be forced, so it's excluded from the TTFS cohort and emits no TTFS row.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved speech-to-text streaming reliability with deterministic close semantics and more accurate timing measurements
  * Enhanced end-of-turn detection and turn finalization handling across providers
  * Optimized metrics calculations for specific model variants

<h3>Greptile Summary</h3>

This PR adds deterministic end-of-speech finalization for AssemblyAI and Deepgram nova models, ensuring both providers emit their final transcript on the VAD end-of-speech signal (ForceEndpoint / Finalize) rather than their own latency-variable turn decisions, aligning them in the TTFS parity cohort. Deepgram Flux is explicitly excluded from the TTFS metric because it offers no way to disable or override its native end-of-turn detection.

- **AssemblyAI**: URL now includes `end_of_turn_confidence_threshold=1.0` (disables semantic auto-finalization), `_send_audio` sends `ForceEndpoint` then waits up to `_FINAL_WAIT_S` for the `Turn(end_of_turn=True)` response before sending `Terminate`, preventing the WS-close race on the final.
- **Deepgram nova**: `endpointing=false` is added to the v1/listen URL for all non-Flux models, `_send_audio` sends `Finalize` then waits for the forced `is_final` before sending `CloseStream`; Flux skips the Finalize block entirely and the orchestrator omits its TTFS row.
- New `asyncio.Event` gate mechanism shared between `_send_audio` and `_receive` in both providers replaces the previous fire-and-forget close pattern.

Safe to merge. The gate mechanism is well-designed, the Flux exclusion is clearly scoped, and the tests cover the critical message-ordering invariants.

The changes add a deterministic close sequence to two providers and a targeted TTFS exclusion in the orchestrator. All three code paths are exercised by new tests, the asyncio.Event gate is correctly initialised and cleared before each forced-final cycle, and no existing metric rows or test assertions are broken. The concerns already noted in the PR thread (server FIFO assumption, FakeWebSocket timing in new tests, and the flux- prefix duplication) are the ceiling on confidence, but none represents a definite defect in the changed paths.

| Filename | Overview |
|----------|----------|
| runner/src/coval_bench/providers/stt/assemblyai.py | Adds ForceEndpoint + asyncio.Event gate before Terminate; end_of_turn_confidence_threshold=1.0 in URL disables semantic auto-finalization; logic is sound, gate clear-before-send pattern is correct for cooperative asyncio. |
| runner/src/coval_bench/providers/stt/deepgram.py | Adds endpointing=false to all non-Flux v1/listen URLs and a Finalize + asyncio.Event gate before CloseStream; Flux send path unchanged; final_event.set() only in the is_final branch, not in TurnInfo/Flux path — correctly aligned with how _send_audio uses the gate. |
| runner/src/coval_bench/runner/orchestrator.py | Adds Flux TTFS exclusion guard; ttfs_value is still computed (no harm) but the Result row is not appended; existing TTFT/AudioToFinal/RTF/WER rows are unaffected. |
| runner/tests/providers/stt/test_assemblyai.py | New test verifies ForceEndpoint precedes Terminate, URL contains threshold param, and audio_to_final_seconds is set; coverage of the gate mechanism is present but the test's final event fires before ForceEndpoint (FakeWebSocket timing), so it primarily validates message ordering rather than end-to-end gate behavior. |
| runner/tests/providers/stt/test_deepgram.py | New tests cover nova Finalize-before-CloseStream ordering, Flux no-Finalize invariant, and endpointing=false in URLs; similar FakeWebSocket timing caveats as AssemblyAI test; endpointing=false is not asserted for default/nova-2 URL (only nova-3). |
| runner/tests/unit/test_orchestrator.py | New test_flux_excluded_from_ttfs correctly validates that TTFS row is absent and TTFT/AudioToFinal/RTF/WER remain; fixture setup with transcript reference ensures WER assertion is sound. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant SA as _send_audio
    participant WS as WebSocket
    participant SRV as Provider Server
    participant RCV as _receive

    Note over SA,RCV: Shared asyncio.Event: final_event

    SA->>WS: audio chunks 1..N
    WS->>SRV: audio chunks
    SRV-->>WS: partial and is_final responses
    WS-->>RCV: messages
    RCV->>RCV: updates last_final_time

    SA->>SA: final_event.clear()
    SA->>WS: ForceEndpoint or Finalize
    WS->>SRV: ForceEndpoint or Finalize
    SRV-->>WS: "Turn(end_of_turn=True) or is_final"
    WS-->>RCV: forced final
    RCV->>RCV: "last_final_time = now"
    RCV->>RCV: final_event.set()

    SA->>SA: "wait_for(final_event, timeout=5s)"
    SA->>WS: Terminate or CloseStream
    WS->>SRV: close

    Note over RCV: audio_to_final_seconds computed from last_final_time
```

<sub>Reviews (3): Last reviewed commit: ["Reset the final-wait event before forcin..."](https://github.com/coval-ai/benchmarks/commit/afc26cb92bef1e1cf37b157f2e35308281f3751a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36601570)</sub>

### DIFF
--- a/runner/src/coval_bench/providers/stt/assemblyai.py
+++ b/runner/src/coval_bench/providers/stt/assemblyai.py
@@ -32,6 +32,10 @@ _SPEECH_MODEL_MAP: dict[str, str] = {
 }
 _WS_BASE = "wss://streaming.assemblyai.com/v3/ws"
 
+# Held high so the semantic end-of-turn model can't auto-finalize mid-utterance
+# before our ForceEndpoint at speech-end (TTFS parity). Tune after a re-run.
+_END_OF_TURN_CONFIDENCE_THRESHOLD = 0.9
+
 
 class AssemblyAIProvider(STTProvider):
     """AssemblyAI v3 streaming STT provider."""
@@ -71,7 +75,10 @@ class AssemblyAIProvider(STTProvider):
 
         try:
             speech_model = _SPEECH_MODEL_MAP[self._model]
-            url = f"{_WS_BASE}?sample_rate={sample_rate}&speech_model={speech_model}"
+            url = (
+                f"{_WS_BASE}?sample_rate={sample_rate}&speech_model={speech_model}"
+                f"&end_of_turn_confidence_threshold={_END_OF_TURN_CONFIDENCE_THRESHOLD}"
+            )
             headers = {"Authorization": self._api_key.get_secret_value()}
             async with ws_client.connect(url, additional_headers=headers) as ws:
                 send_task = asyncio.create_task(
@@ -117,6 +124,8 @@ class AssemblyAIProvider(STTProvider):
                     first_chunk = False
                 await ws.send(chunk)
                 await asyncio.sleep(realtime_resolution)
+            # Force finalization at speech-end (TTFS parity) before terminating.
+            await ws.send(json.dumps({"type": "ForceEndpoint"}))
             await ws.send(json.dumps({"type": "Terminate"}))
         except Exception as exc:
             logger.exception("assemblyai send error", error=str(exc))
--- a/runner/src/coval_bench/providers/stt/assemblyai.py
+++ b/runner/src/coval_bench/providers/stt/assemblyai.py
@@ -12,6 +12,7 @@ Close: {"type": "Terminate"}
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import json
 import time  # monotonic clock — wall-clock can step on NTP sync
 from typing import Any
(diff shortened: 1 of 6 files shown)


coval-cale (approved):
Nice!

PR #100, opened 2026-06-11
Per window stats matviews

coval-cale:
This adds materialized views to speed up data loading.
- Before: Every request for aggregated stats combed through the entire db and built the result in realtime.
- After: Adds three tables to the db that can be updated when new data hits. They just store all the stats for our predefined windows, making lookup blazingly fast.

Note: This doesn't address the timeline view lag, just the stats aggregated over a window.
Note: This does nothing for now. A separate PR to the infra repo will create a hook to update this whenever new benchmarks land. 

## Summary by CodeRabbit

* **New Features**
  * Extended aggregate statistics now include additional percentiles: p90, p95, and p99.
  * Leaderboard and aggregates endpoints now support multiple time windows: 24h, 7d, and 30d.

* **Refactor**
  * Heatmap visualization updated to display absolute latency percentiles and inter-quartile ranges instead of comparative deltas.

<h3>Greptile Summary</h3>

This PR replaces live aggregation queries in the `/results/aggregates` and `/leaderboard` endpoints with reads from three new PostgreSQL materialized views (`results_24h`, `results_7d`, `results_30d`), each pre-computing wide per-model stats (avg, stddev, p25–p99, min, max, count) for their respective time window. The frontend heatmap is simplified by unifying the separate STT-delta and TTS-absolute paths into a single absolute-percentile implementation.

- **Backend**: Alembic migration drops the old `results_24h` and creates all three views with unique indexes (enabling `CONCURRENTLY` refresh); `WINDOW_VIEWS` dict and `_STATS_SQL_TEMPLATE`/`_MV_SQL_TEMPLATE` string templates safely inject view names from validated Python constants; the `series` block in aggregates remains a live query.
- **Frontend**: `useChartData` removes the delta-relative-to-fastest STT heatmap computation and the separate `ttsHeatmapDataMemo`, replacing both with a single `heatmapDataMemo` that reads absolute percentiles via `toDisplayUnits()` for both tabs; `HeatmapPlot` metric labels are now tab-independent.

Safe to merge; the core view-lookup and SQL template logic is correct and the view name interpolation is properly guarded by validated Python constants.

The migration, API routing changes, schema additions, and test updates are all consistent with each other. The main items to be aware of are the downgrade path leaving a non-unique index (which would prevent CONCURRENTLY refresh if the cron is updated before a rollback) and the test conftest duplicating the migration SQL rather than sharing it. Neither blocks a forward deployment, but both could cause pain in a rollback or future migration scenario.

runner/tests/api/conftest.py (duplicated migration SQL) and the downgrade block in 20260611_0005_per_window_stats_matviews.py (non-unique index after rollback).

| Filename | Overview |
|----------|----------|
| runner/src/coval_bench/db/migrations/versions/20260611_0005_per_window_stats_matviews.py | New migration creates three materialized views (results_24h/7d/30d) with unique indexes for CONCURRENTLY refresh; downgrade recreates results_24h without a unique index, which could break cron-based CONCURRENTLY refresh after a rollback. |
| runner/src/coval_bench/api/routers/aggregates.py | model_stats block now reads from the validated-key matview template instead of a live aggregation; view name interpolation is safe because window is constrained by WindowLiteral and WINDOW_VIEWS is a hardcoded dict. |
| runner/src/coval_bench/api/routers/leaderboard.py | All three windows now serve from their respective matviews via a single SQL template; live aggregation path removed; view name interpolation is safe. |
| runner/src/coval_bench/api/common.py | Adds WINDOW_VIEWS dict mapping validated WindowLiteral keys to schema-qualified view names; comment clarification on WINDOW_INTERVALS usage. |
| runner/src/coval_bench/api/schemas.py | ModelStatEntry gains p90, p95, p99 fields to match the wider matview shape; no schema compatibility issues. |
| runner/tests/api/conftest.py | Test schema setup now creates all three matviews by duplicating the migration SQL inline; duplication creates a maintenance risk if the migration SQL changes without updating the conftest. |
| runner/tests/api/test_aggregates.py | Tests correctly call _refresh_mv before asserting on model_stats; new p90/p95/p99 assertions added with correct expected values. |
| web/hooks/useChartData.ts | Unifies STT and TTS heatmap paths into a single memo using absolute percentiles with toDisplayUnits(); removes the delta-relative-to-fastest STT computation and the separate ttsHeatmapDataMemo; TTS heatmap now applies toDisplayUnits() where the old code did not. |
| web/hooks/useDashboardState.tsx | Removes the page-conditional heatmap dispatch; now calls the single getHeatmapData() for both pages. |
| web/components/charts/d3/HeatmapPlot.tsx | metrics array is now static (no tab-conditional Delta vs Latency labels); dependency on activeTab removed from the useMemo; always shows absolute latency percentile labels. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client
    participant API
    participant MatView as Materialized View<br/>(results_24h/7d/30d)
    participant ResultsDB as results / runs tables
    participant Cron as Cloud Scheduler Cron

    Client->>API: "GET /v1/results/aggregates?window=24h"
    API->>MatView: "SELECT provider, model, … FROM results_24h WHERE benchmark=…"
    MatView-->>API: model_stats rows (pre-aggregated)
    API->>ResultsDB: SELECT … FROM results JOIN runs … (live series query)
    ResultsDB-->>API: series rows
    API-->>Client: "{model_stats, series}"

    Client->>API: "GET /v1/leaderboard?window=7d"
    API->>MatView: "SELECT … FROM results_7d WHERE metric_type=… AND benchmark=…"
    MatView-->>API: leaderboard entries
    API-->>Client: "{entries}"

    Cron->>MatView: REFRESH MATERIALIZED VIEW CONCURRENTLY results_24h
    Cron->>MatView: REFRESH MATERIALIZED VIEW CONCURRENTLY results_7d
    Cron->>MatView: REFRESH MATERIALIZED VIEW CONCURRENTLY results_30d
```

<details><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `runner/tests/api/conftest.py`, line 380-405 ([link](https://github.com/coval-ai/benchmarks/blob/8e849cdfebbf06ec33e84737990dbd4fdb2fe8ee/runner/tests/api/conftest.py#L380-L405)) 

   <a href="#">P2</a> **Test schema duplicates migration SQL verbatim**

   The materialized-view DDL in `conftest.py` (`_apply_schema`) is an exact copy of the SQL from `20260611_0005_per_window_stats_matviews.py`. Both the migration and the test fixture will now need to be updated in lock-step for any future change to the view shape (e.g. adding a column, changing the `runs` JOIN condition). If the migration is updated without updating the conftest, tests will silently exercise a view definition that no longer matches production, masking real regressions.

2. `runner/src/coval_bench/db/migrations/versions/20260611_0005_per_window_stats_matviews.py`, line 320-341 ([link](https://github.com/coval-ai/benchmarks/blob/8e849cdfebbf06ec33e84737990dbd4fdb2fe8ee/runner/src/coval_bench/db/migrations/versions/20260611_0005_per_window_stats_matviews.py#L320-L341)) 

   <a href="#">P2</a> **Downgrade recreates `results_24h` without a unique index**

   The `upgrade()` path creates each view with a `UNIQUE INDEX` (`{name}_group_key`) that allows `REFRESH MATERIALIZED VIEW CONCURRENTLY`. The `downgrade()` path restores `results_24h` with only the original non-unique `results_24h_lookup_idx`. If the companion infra PR updates the Cloud Scheduler cron to use `CONCURRENTLY` refresh for all three views (which the unique indexes are there to support), a rollback will leave the cron unable to refresh `results_24h` until a unique index is added manually.

   Consider adding a unique index on `(provider, model, benchmark, metric_type)` in the downgrade path, or documenting that the cron must be changed back to a non-concurrent refresh before running the downgrade.

3. `web/hooks/useChartData.ts`, line 299-309 ([link](https://github.com/coval-ai/benchmarks/blob/8e849cdfebbf06ec33e84737990dbd4fdb2fe8ee/web/hooks/useChartData.ts#L299-L309)) 

   <a href="#">P2</a> **TTS heatmap now applies `toDisplayUnits()` — behaviour change**

   The old `ttsHeatmapDataMemo` used raw `latencyStat.p25/p50/p75` values without calling `toDisplayUnits()`. The unified `heatmapDataMemo` now applies `toDisplayUnits(latencyStat.pXX)` for both tabs. All other TTS charts (box plot, scatter) already go through `toDisplayUnits`, so this is almost certainly correcting a pre-existing inconsistency. Worth explicitly verifying that the TTS latency values stored in the database are not already in display units — if they are, the heatmap will now show doubled/scaled values compared to what was shown before this PR.

</details>

<sub>Reviews (1): Last reviewed commit: ["Show absolute latency percentiles in the..."](https://github.com/coval-ai/benchmarks/commit/8e849cdfebbf06ec33e84737990dbd4fdb2fe8ee) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37080209)</sub>

### DIFF
--- a/runner/src/coval_bench/db/migrations/versions/20260611_0005_per_window_stats_matviews.py
+++ b/runner/src/coval_bench/db/migrations/versions/20260611_0005_per_window_stats_matviews.py
@@ -1,0 +1,98 @@
+# Copyright 2026 The Coval Benchmarks Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Per-window stats materialized views: results_24h, results_7d, results_30d.
+
+One view per dashboard window, all sharing a wide stats shape per
+(provider, model, benchmark, metric_type): avg, sample stddev, min,
+p25/p50/p75/p90/p95/p99 (a single PERCENTILE_CONT over an array, so one
+sort), max, count. Rows gate on result status='success', a non-null metric_value,
+and a parent run in (succeeded, partial) — the same filters the aggregates
+endpoint applied live.
+
+results_24h is dropped and recreated in this shape (it previously held only
+avg/p50/p95/n and did not gate on the parent run). Each view gets a unique
+index on the group key so it can be refreshed with REFRESH MATERIALIZED VIEW
+CONCURRENTLY. Refresh remains out-of-band (Cloud Scheduler cron); this
+migration only creates the views, so they hold data as of migration time
+until the first refresh.
+"""
+
+from __future__ import annotations
+
+from alembic import op
+
+revision = "20260611_0005"
+down_revision = "20260605_0004"
+branch_labels = None
+depends_on = None
+
+_WINDOWS: dict[str, str] = {
+    "results_24h": "24 hours",
+    "results_7d": "7 days",
+    "results_30d": "30 days",
+}
+
+
+def _view_sql(name: str, interval: str) -> str:
+    # S608 false-positive: name and interval come from the _WINDOWS constant.
+    return f"""
+        CREATE MATERIALIZED VIEW benchmarks_v2.{name} AS
+        SELECT provider, model, benchmark, metric_type,
+               avg_value, stddev_value, min_value,
+               pct[1] AS p25, pct[2] AS p50, pct[3] AS p75,
+               pct[4] AS p90, pct[5] AS p95, pct[6] AS p99,
+               max_value, sample_count
+        FROM (
+            SELECT r.provider, r.model, r.benchmark, r.metric_type,
+                   AVG(r.metric_value)::float8 AS avg_value,
+                   COALESCE(STDDEV_SAMP(r.metric_value), 0)::float8 AS stddev_value,
+                   MIN(r.metric_value)::float8 AS min_value,
+                   PERCENTILE_CONT(ARRAY[0.25, 0.5, 0.75, 0.9, 0.95, 0.99])
+                       WITHIN GROUP (ORDER BY r.metric_value)::float8[] AS pct,
+                   MAX(r.metric_value)::float8 AS max_value,
+                   COUNT(*)::int AS sample_count
+            FROM benchmarks_v2.results r
+            JOIN benchmarks_v2.runs rn ON rn.id = r.run_id
+            WHERE r.status = 'success'
+              AND rn.status IN ('succeeded', 'partial')
+              AND r.metric_value IS NOT NULL
+              AND r.created_at >= now() - INTERVAL '{interval}'
+            GROUP BY r.provider, r.model, r.benchmark, r.metric_type
+        ) stats
+    """  # noqa: S608
+
+
+def upgrade() -> None:
+    """Recreate results_24h in the wide shape and add results_7d/results_30d."""
+    op.execute("DROP MATERIALIZED VIEW IF EXISTS benchmarks_v2.results_24h")
+    for name, interval in _WINDOWS.items():
+        op.execute(_view_sql(name, interval))
+        op.execute(
+            f"CREATE UNIQUE INDEX {name}_group_key "
+            f"ON benchmarks_v2.{name} (provider, model, benchmark, metric_type)"
+        )
+
+
+def downgrade() -> None:
+    """Drop the per-window views and restore the original results_24h."""
+    for name in _WINDOWS:
+        op.execute(f"DROP MATERIALIZED VIEW IF EXISTS benchmarks_v2.{name}")
+    op.execute(
+        """
+        CREATE MATERIALIZED VIEW benchmarks_v2.results_24h AS
+        SELECT provider, model, benchmark, metric_type,
+               avg(metric_value)                                        AS avg_value,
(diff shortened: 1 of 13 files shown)


runner/src/coval_bench/db/migrations/versions/20260611_0005_per_window_stats_matviews.py
```diff
@@ -0,0 +1,98 @@
+# Copyright 2026 The Coval Benchmarks Authors
+# SPDX-License-Identifier: Apache-2.0
+
```
seribaymadina, line 3: is the infra PR that adds refresh for 7 and 30 days to the cron landing before this? clarifying cuz might get frozen if we forget to do it

seribaymadina (approved):

runner/src/coval_bench/db/migrations/versions/20260611_0005_per_window_stats_matviews.py
coval-cale: Gonna land right after! Nice catch!

PR #108, opened 2026-06-12
Add series rollup table (results_by_bucket)

coval-cale:
The timeline series still ran the live query over raw results on every cache miss (a 500k+ row scan on 30d). This adds a `results_by_bucket` rollup keyed `(provider, model, benchmark, metric_type, bucket_at)` holding min/p25/p50/p75/max/value_sum/sample_count, and serves the aggregates series block from it as an index range read over `(benchmark, bucket_at)`. No runs join, no GROUP BY over raw rows, and one table serves every window.

The orchestrator's end-of-run hook fills a run's bucket next to the existing matview refresh: delete-then-insert recomputed from raw rows, so re-runs sharing a slot and double-fired hooks converge to the same state. Failed runs are skipped. The migration creates the table and backfills it from existing results.

Latency timelines now plot the bucket median (p50); WER plots value_sum / sample_count.

Deploy notes:
- Prod migration is manual: `coval-bench db migrate` after the image bump.
- Migrations deploy separately from images, so rerun the backfill (truncate + the migration insert-select) once the new runner image is live to cover the gap.

Tests: migration up/down + backfill over pre-existing rows, refresh_bucket idempotency and shared-slot recompute, failed-run exclusion, orchestrator hook firing on succeeded/partial/SIGTERM-partial and not on failure, and the reshaped series assertions in the aggregates API tests.

## Summary by CodeRabbit

* **New Features**
  * Aggregates endpoint now returns pre-aggregated per-bucket series with full distribution fields (min, p25, p50, p75, max), bucketed value sums and sample counts; schema updated accordingly.

* **Improvements**
  * Endpoint reads precomputed rollups for faster, consistent series results.
  * Timeline charts use medians for latency-like metrics and weighted means for WER.
  * Series rollups are refreshed after successful/partial runs (including SIGTERM-partial) to improve freshness.

* **Database Migration**
  * Adds a rollup table to store per-bucket aggregates for series reads.

* **Tests**
  * Expanded tests to cover migration backfill, rollup refresh behavior, and updated series assertions.

<h3>Greptile Summary</h3>

This PR replaces the live `GROUP BY` query over 500k+ raw rows in the aggregates `series` block with a pre-aggregated `results_by_bucket` rollup table, keyed on `(provider, model, benchmark, metric_type, bucket_at)` and holding full distribution columns (min/p25/p50/p75/max/value_sum/sample_count). The orchestrator's end-of-run hook populates it via a delete-then-recompute-in-one-transaction approach serialised per bucket with a PostgreSQL advisory lock, making simultaneous refreshes idempotent.

- **New rollup table + migration**: `results_by_bucket` is created with a primary key and a `(benchmark, bucket_at)` index; the migration backfills the table from all existing results, and the `downgrade` path cleanly drops the table.
- **Writer & orchestrator hook**: `RunWriter.refresh_bucket` runs a transaction-scoped advisory lock → DELETE → INSERT from raw rows for the entire time-bucket, and `_refresh_series_bucket` wraps it with a 3-attempt retry. The orchestrator calls it after `SUCCEEDED`/`PARTIAL` runs (including SIGTERM-partial), skipping `FAILED` runs.
- **Frontend**: `useChartData` is updated to serve p50 for latency metrics and `value_sum / sample_count` for WER, with a forward-compatible fallback to `avg_value` for pre-rollup API responses.

Safe to merge. The rollup table write path is correctly transactional, advisory-locked per bucket, and idempotent; the API read path is a straightforward index-range scan.

The delete-then-recompute strategy is sound: the advisory lock serialises concurrent refreshes, DELETE and INSERT share a single transaction so readers never see a partially-refreshed bucket, and the retry wrapper prevents transient errors from failing the run. Tests cover migration, idempotency, shared-slot recompute, failed-run exclusion, and orchestrator hook firing.

No files require special attention beyond the already-flagged test gaps in runner/tests/api/test_aggregates.py and runner/tests/api/conftest.py.

| Filename | Overview |
|----------|----------|
| runner/src/coval_bench/db/writer.py | Adds refresh_bucket with advisory lock, DELETE + INSERT in one transaction. Early-return on NULL scheduled_at is safe. |
| runner/src/coval_bench/runner/orchestrator.py | Adds _refresh_series_bucket with 3-attempt retry, wired into normal and SIGTERM-partial end-of-run paths. |
| runner/src/coval_bench/db/migrations/versions/20260611_0006_results_by_bucket.py | Creates results_by_bucket with correct PK and index, backfills from raw rows. Only integer constant interpolated in SQL. |
| runner/src/coval_bench/api/routers/aggregates.py | Replaces live GROUP-BY with indexed range read from results_by_bucket. Cache-key structure unchanged. |
| runner/tests/api/test_aggregates.py | Updates series assertions to check full distribution shape. Window filter not exercised in test_window_excludes_old_rows. |
| runner/tests/api/conftest.py | Adds results_by_bucket DDL and _fill_buckets helper. TRUNCATE + INSERT are two separate autocommit ops (not atomic). |

</details>

<details><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `runner/tests/api/test_aggregates.py`, line 164-176 ([link](https://github.com/coval-ai/benchmarks/blob/a98620253089d91273c4f3b44a5e5d80f8878668/runner/tests/api/test_aggregates.py#L164-L176)) 

   <a href="#">P2</a> **Series time-window filter is not exercised after the rollup change**

   `test_window_excludes_old_rows` doesn't call `_fill_buckets`, so `results_by_bucket` is empty and the `series` block is vacuously `[]` for every window. No test inserts a row with a `bucket_at` that falls outside the 24h window and then asserts it is absent from `series` (but present in the 30d window). The `_SERIES_SQL` filter `bucket_at >= NOW() - %(interval)s::interval` is the key correctness predicate for the new path, so it's worth a dedicated assertion alongside the existing `model_stats` check.

</details>

<sub>Reviews (3): Last reviewed commit: ["Serialize bucket refreshes and retry tra..."](https://github.com/coval-ai/benchmarks/commit/d9fc91a1ae68ad93b38523490e7bccde55513f66) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37300243)</sub>

### DIFF
--- a/runner/src/coval_bench/db/writer.py
+++ b/runner/src/coval_bench/db/writer.py
@@ -141,6 +141,70 @@ class RunWriter:
                 await cur.executemany(sql, params)
             await conn.commit()
 
+    async def refresh_bucket(self, run_id: int, *, period_seconds: int) -> None:
+        """Recompute the series rollup bucket for this run's scheduled_at slot.
+
+        Delete-then-insert the whole bucket from raw result rows in one
+        transaction. Recomputing the full bucket (not just this run's rows)
+        keeps it correct when runs share a slot, and makes the call idempotent.
+        Runs without a ``scheduled_at`` are skipped — the migration backfill
+        owns those.
+        """
+        async with self._pool.connection() as conn:
+            async with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+                await cur.execute(
+                    "SELECT scheduled_at FROM benchmarks_v2.runs WHERE id = %s",
+                    (run_id,),
+                )
+                row = await cur.fetchone()
+                bucket_at = row["scheduled_at"] if row is not None else None
+                if bucket_at is None:
+                    return
+
+                # Two statements on purpose: a data-modifying CTE
+                # (WITH deleted AS (DELETE ...) INSERT) collides on the primary
+                # key — the INSERT cannot see the CTE's deletes.
+                params = {"bucket": bucket_at, "period": period_seconds}
+                await cur.execute(
+                    "DELETE FROM benchmarks_v2.results_by_bucket WHERE bucket_at = %(bucket)s",
+                    params,
+                )
+                # Bucket membership: scheduled_at matches exactly, or a legacy
+                # null-scheduled row whose created_at falls in
+                # [bucket_at, bucket_at + period).
+                await cur.execute(
+                    """
+                    INSERT INTO benchmarks_v2.results_by_bucket
+                        (provider, model, benchmark, metric_type, bucket_at,
+                         min_value, p25, p50, p75, max_value, value_sum, sample_count)
+                    SELECT r.provider, r.model, r.benchmark, r.metric_type, %(bucket)s,
+                           MIN(r.metric_value)::float8,
+                           PERCENTILE_CONT(0.25) WITHIN GROUP (ORDER BY r.metric_value)::float8,
+                           PERCENTILE_CONT(0.5)  WITHIN GROUP (ORDER BY r.metric_value)::float8,
+                           PERCENTILE_CONT(0.75) WITHIN GROUP (ORDER BY r.metric_value)::float8,
+                           MAX(r.metric_value)::float8,
+                           SUM(r.metric_value)::float8,
+                           COUNT(*)::int
+                    FROM benchmarks_v2.results r
+                    JOIN benchmarks_v2.runs rn ON rn.id = r.run_id
+                    WHERE r.status = 'success'
+                      AND rn.status IN ('succeeded', 'partial')
+                      AND r.metric_value IS NOT NULL
+                      AND (
+                          rn.scheduled_at = %(bucket)s
+                          OR (
+                              rn.scheduled_at IS NULL
+                              AND r.created_at >= %(bucket)s
+                              AND r.created_at < %(bucket)s
+                                  + (%(period)s::double precision) * INTERVAL '1 second'
+                          )
+                      )
+                    GROUP BY r.provider, r.model, r.benchmark, r.metric_type
+                    """,
+                    params,
+                )
+            await conn.commit()
+
     async def finish_run(
         self,
         run_id: int,
--- a/runner/src/coval_bench/db/writer.py
+++ b/runner/src/coval_bench/db/writer.py
@@ -145,10 +145,10 @@ class RunWriter:
         """Recompute the series rollup bucket for this run's scheduled_at slot.
 
         Delete-then-insert the whole bucket from raw result rows in one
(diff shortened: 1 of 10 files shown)


runner/src/coval_bench/db/writer.py
```diff
@@ -141,6 +141,70 @@ async def record_results(self, results: Sequence[Result]) -> None:
...
+                bucket_at = row["scheduled_at"] if row is not None else None
+                if bucket_at is None:
+                    return
+
+                # Two statements on purpose: a data-modifying CTE
+                # (WITH deleted AS (DELETE ...) INSERT) collides on the primary
+                # key — the INSERT cannot see the CTE's deletes.
+                params = {"bucket": bucket_at, "period": period_seconds}
+                await cur.execute(
+                    "DELETE FROM benchmarks_v2.results_by_bucket WHERE bucket_at = %(bucket)s",
```
seribaymadina, line 179: seems like could be an issue where on an empty bucket due to DELETE not locking things, two transactions for the same slot both would INSERT -> could silently drop the run from the slot with no later refresh

runner/src/coval_bench/runner/orchestrator.py
```diff
@@ -648,6 +648,18 @@ def _emit_posthog(client: Posthog | None, event: str, properties: dict[str, Any]
         _log.warning("posthog_emit_failed", event_name=event, exc_info=True)
 
 
+async def _refresh_series_bucket(writer: Any, run_id: int, settings: Settings) -> None:  # noqa: ANN401 — RunWriter, lazy-imported by the caller
+    """Best-effort refresh of the run's series rollup bucket; never raises.
+
+    A failure leaves only this bucket stale (later runs fill their own
+    buckets, not this one) — not worth failing the run.
```
seribaymadina: do you think in the future we might consider failed bucket refresh auto recover?

runner/src/coval_bench/db/writer.py
seribaymadina: feel free to push bac if you disagree!

seribaymadina (changes requested):

runner/src/coval_bench/runner/orchestrator.py
seribaymadina: such an awesome work done with this though, again 50% increase in page load time is huge!

runner/src/coval_bench/db/writer.py
coval-cale: Nice catch! I had this exactly opposite in my head. Thanks so much!

seribaymadina (approved):
lgtm!!

PR #117, opened 2026-06-15
Handle empty xAI transcripts and exclude Grok from TTFT

seribaymadina:
An empty `transcript.done` no longer stamps a finalization time: the session is failed (and so retryable) instead of recording a spurious ~5s TTFS/AudioToFinal success.

Grok's TTFT is a fixed min-audio floor (measured ~1.1s, stdev 0.05s, uncorrelated with clip length), not engine speed, so it is omitted the way Flux is excluded from TTFS. Grok ranks on TTFS and WER.

Tests: `test_xai_empty_session_errors` (+ fixture) and `test_xai_excluded_from_ttft`.

## Summary by CodeRabbit

* **Bug Fixes**
  * xAI STT provider now properly detects empty transcript sessions and returns an error message instead of proceeding with invalid metrics.
  * xAI STT provider excluded from Time-To-First-Token (TTFT) metric reporting.

* **Tests**
  * Added test coverage for empty transcript session handling.
  * Added validation for xAI TTFT exclusion from metrics.

<h3>Greptile Summary</h3>

This PR fixes two related issues with the xAI (`grok-stt`) provider: an empty `transcript.done` no longer stamps a spurious `audio_to_final_seconds`, and TTFT is excluded from the metrics table because Grok's min-audio gate produces a measurement that is an artifact of buffering, not engine speed. The model registry is also updated from `_RETIRED` to `_ACTIVE`.

- **xai.py**: The `last_final_time` assignment is moved inside the `if done_transcript:` block so it only updates on non-empty terminal events; after `finalize_transcript`, a new early-return guard checks `complete_transcript is None` and sets an error instead of recording `audio_to_final_seconds` for a dead session.
- **orchestrator.py**: A `ttft_excluded = entry.provider == "xai"` guard skips appending the TTFT `Result` row, mirroring the existing Flux/TTFS exclusion pattern.
- **Tests**: `test_xai_empty_session_errors` (with fixture) and `test_xai_excluded_from_ttft` provide direct coverage of both new behaviors.

Safe to merge — both changes are narrowly scoped, well-tested, and do not alter logic for any provider other than xAI.

The empty-session guard is a straightforward early-return after finalize_transcript, and the TTFT exclusion mirrors the established Flux/TTFS pattern. New tests cover both paths directly, and the registry promotion is consistent with the companion test update in test_providers.py.

| Filename | Overview |
|----------|----------|
| runner/src/coval_bench/providers/stt/xai.py | Indentation fix moves last_final_time stamp inside the non-empty done-text branch; new guard after finalize_transcript returns early with an error when complete_transcript is None, preventing spurious audio_to_final on empty sessions. |
| runner/src/coval_bench/runner/orchestrator.py | Adds ttft_excluded = entry.provider == 'xai' guard to skip appending the TTFT Result row for xAI, consistent with the Flux/TTFS exclusion pattern. |
| runner/src/coval_bench/registries/models.py | grok-stt status promoted from _RETIRED to _ACTIVE; the companion test in test_providers.py is updated to assert disabled=False. |
| runner/tests/providers/stt/test_xai.py | New test_xai_empty_session_errors verifies error is set and no timing metrics are recorded when the session produces no speech. |
| runner/tests/unit/test_orchestrator.py | New test_xai_excluded_from_ttft confirms TTFT row is absent while AudioToFinal, RTF, TTFS, and WER are still emitted for the xAI provider. |
| runner/tests/providers/stt/fixtures/xai/events-empty-session.json | New fixture with transcript.created + transcript.done with empty text, sufficient to exercise the dead-session error path. |
| runner/tests/api/test_providers.py | Assertion updated to expect disabled=False for grok-stt following its _RETIRED → _ACTIVE promotion. |

</details>

<h3>Flowchart</h3>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[transcript.done received] --> B{done_transcript non-empty?}
    B -- Yes --> C[set_first_token / add_partial / append to final_segments]
    C --> D[last_final_time = now]
    D --> E[break]
    B -- No --> E

    E --> F[finalize_transcript]
    F --> G{complete_transcript is None?}
    G -- Yes --> H[result.error = 'xAI returned an empty transcript']
    H --> I[return — no timing metrics stamped]
    G -- No --> J{last_final_time AND audio_start_time set?}
    J -- Yes --> K[audio_to_final_seconds = last_final_time - audio_start_time]
    J -- No --> L[audio_to_final_seconds stays None]

    subgraph Orchestrator
        M[entry.provider == 'xai'?] -- Yes --> N[skip TTFT Result row]
        M -- No --> O[append TTFT Result row]
    end
```

<sub>Reviews (3): Last reviewed commit: ["Enable Grok STT in the model registry"](https://github.com/coval-ai/benchmarks/commit/2a4fa7f130466000a8796fc0e5c9b093c5deb7f8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37728444)</sub>

### DIFF
--- a/runner/src/coval_bench/runner/orchestrator.py
+++ b/runner/src/coval_bench/runner/orchestrator.py
@@ -274,25 +274,29 @@ async def _run_stt_item(
             transcription_result.complete_transcript if transcription_result else None
         )
 
-        # 1. TTFT
+        # 1. TTFT — time-to-first-partial from first audio. For xAI Grok this reflects
+        # its min-audio gate, not engine speed, so it misrepresents the provider; TTFS
+        # (re-anchored at end-of-speech) is the fair metric. Omit TTFT for xAI; the rest run.
         ttft_status, ttft_error = _metric_outcome(
             ttft_seconds, item_error, Metric.TTFT, ResultStatus
         )
-        results.append(
-            Result(
-                run_id=run_id,
-                provider=entry.provider,
-                model=entry.model,
-                benchmark=Benchmark.STT,
-                metric_type=Metric.TTFT,
-                metric_value=ttft_seconds,
-                metric_units=METRIC_SPECS[Metric.TTFT].units,
-                audio_filename=audio_path.name,
-                transcript=complete_transcript,
-                status=ttft_status,
-                error=ttft_error,
+        ttft_excluded = entry.provider == "xai"
+        if not ttft_excluded:
+            results.append(
+                Result(
+                    run_id=run_id,
+                    provider=entry.provider,
+                    model=entry.model,
+                    benchmark=Benchmark.STT,
+                    metric_type=Metric.TTFT,
+                    metric_value=ttft_seconds,
+                    metric_units=METRIC_SPECS[Metric.TTFT].units,
+                    audio_filename=audio_path.name,
+                    transcript=complete_transcript,
+                    status=ttft_status,
+                    error=ttft_error,
+                )
             )
-        )
 
         # 2. AudioToFinal
         atf_status, atf_error = _metric_outcome(
--- a/runner/src/coval_bench/providers/stt/xai.py
+++ b/runner/src/coval_bench/providers/stt/xai.py
@@ -200,8 +200,8 @@ class XaiSTTProvider(STTProvider):
                         set_first_token(result, done_transcript, now=now)
                         add_partial_transcript(result, done_transcript)
                         final_segments.append(done_transcript)
-                    if result.audio_start_time is not None:
-                        last_final_time = now
+                        if result.audio_start_time is not None:
+                            last_final_time = now
                     break
 
                 if event_type == "error":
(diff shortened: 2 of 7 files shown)


runner/src/coval_bench/runner/orchestrator.py
```diff
@@ -274,25 +274,29 @@ async def _run_stt_item(
...
-                model=entry.model,
-                benchmark=Benchmark.STT,
-                metric_type=Metric.TTFT,
-                metric_value=ttft_seconds,
-                metric_units=METRIC_SPECS[Metric.TTFT].units,
-                audio_filename=audio_path.name,
-                transcript=complete_transcript,
-                status=ttft_status,
-                error=ttft_error,
+        ttft_excluded = entry.provider == "xai"
```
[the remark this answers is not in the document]
seribaymadina: it's alright, we will consider a method that will keep things fair for every model

coval-cale, line 283: Might be worth abstracting this + the deepgram models in to something like a metric eligibility registry. That way we can flip things on and off from a single file instead of sprinkling these through the codebase.

coval-cale (approved):
We should do a registry for this stuff (I think) but that can be a separate pr. Nice work!

runner/src/coval_bench/runner/orchestrator.py
seribaymadina: I am thinking that there might be a way for us to include them back to TTFT in the future, for now leaving it. thanks for catching it!!

PR #127, opened 2026-06-16
Add OpenAI gpt-4o-transcribe and gpt-4o-mini-transcribe STT

coval-cale:
Both run on the existing realtime websocket transcription path alongside gpt-realtime-whisper.


callumreid (approved):
pop off, big dog

PR #210, opened 2026-07-01
Gate arena TTS providers on mounted API keys

seribaymadina:
A new TTS provider can be marked arena-active in this repo without its API key being mounted on benchmarks-api in benchmark-infra — today that only surfaces as a failed live battle. This catches the gap at PR time.

An `arena_enabled` flag on `RegisteredModel` decouples arena membership from dashboard `status` (default `True`, so every active TTS model is in the arena unless explicitly opted out — inworld and groq are opted out here). A single `PROVIDER_ENV` map is the source of truth for provider→key naming, guarded by tests that fail if an arena provider has no entry or a name drifts from `Settings`. A path-filtered workflow runs only when the TTS provider registry changes: it reads what benchmark-infra mounts on the Cloud Run service and fails the PR if an arena-eligible provider's key isn't there, naming exactly where to add it.

There is deliberately no runtime credential filter. A missing key still surfaces as a 502 on the live battle path (unchanged behavior) — the CI check is the mitigation, catching the missing mount at PR time rather than in a live battle. It's a visible signal on provider PRs today, not a required merge gate; making it hard-block would need a ruleset change.

<h3>Greptile Summary</h3>

This PR adds an `arena_enabled` flag to `RegisteredModel` to decouple arena membership from dashboard `status`, introduces `PROVIDER_ENV` as the single source of truth for provider → API-key name mapping, and wires a new path-filtered CI workflow that cross-checks the arena roster against what the benchmark-infra Cloud Run service actually mounts.

- `active_tts_models()` is narrowed to require `arena_enabled=True`, three `ACTIVE` models that lack mounted keys are explicitly opted out, and a new `provider_keys.py` module maps every provider to its expected env-var name.
- `check_arena_keys.py` parses the infra Terraform file (fetched via `gh api`) and fails if any arena-eligible provider's key is absent; the workflow degrades gracefully to a warning (not an error) when the infra token is unavailable so fork PRs are not hard-blocked.
- Unit tests guard against `PROVIDER_ENV` drift from both the active roster and `Settings` field names.

Safe to merge with awareness that the seed path and runtime filtering gaps noted in earlier review threads remain open.

The structural changes are well-contained and the new tests provide good coverage of the registry/key-map invariants. The two issues flagged in earlier review rounds — seed_demo_battles calling select_pair without first checking that the roster has at least two models, and active_tts_models having no runtime key-based filtering despite the PR description promising it — are still unresolved in the current code, and both affect live arena paths when keys are missing.

runner/src/coval_bench/arena/seed.py and runner/src/coval_bench/arena/pairing.py carry the open issues from the earlier review; the new files (provider_keys.py, check_arena_keys.py, the workflow) look correct.

<sub>Reviews (5): Last reviewed commit: ["Gate arena TTS providers on mounted API ..."](https://github.com/coval-ai/benchmarks/commit/1030d6263b0b286c23b8b5c6b4f89b42e0a7c9ff) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41200956)</sub>

### DIFF
--- a/.github/workflows/check-arena-keys.yml
+++ b/.github/workflows/check-arena-keys.yml
@@ -1,0 +1,72 @@
+# Copyright 2026 The Coval Benchmarks Authors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Cross-repo parity gate: when a PR changes the TTS provider registry, every
+# arena-eligible provider's API key must be mounted on the benchmarks-api Cloud
+# Run service in coval-ai/benchmark-infra. Path-filtered so it only runs on
+# provider-addition PRs — never on unrelated branches or already-merged work.
+name: Arena key parity
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - "runner/src/coval_bench/registries/models.py"
+      - "runner/src/coval_bench/registries/provider_keys.py"
+      - "runner/src/coval_bench/providers/tts/**"
+
+permissions:
+  contents: read
+
+jobs:
+  parity:
+    name: Arena provider keys mounted in infra
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v7
+        with:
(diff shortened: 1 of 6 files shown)


coval-cale (approved):
Heck ya!

PR #214, opened 2026-07-02
Add standardized dataset builder framework

seribaymadina:
One shared pipeline for building benchmark datasets — clean, balance, transcode, hash, manifest, upload — where only the per-dataset source (download + parse) is written each time.

`balanced_sample` is the core: dedup, per-dimension tag-coverage checks (warn <0.95, abort <0.50), exclude untagged clips rather than default them, then a stratified round-robin so each balance dimension is even.

A generic Hugging Face source builds a manifest from just a dataset path — resolve splits, auto-detect columns, metadata-first paging, and scaffold a handwritten adapter when a dataset can't be read. SLURP `s2s-v1` is the first adapter; `coval-build-dataset` is the CLI.

Experimental. Follow-ups: parquet fallback for unconverted datasets, cache-management flags, stt-v1 migration.

<h3>Greptile Summary</h3>

This PR introduces a shared pipeline for building benchmark datasets — `framework.py` owns the clean → balance → transcode → hash → manifest → upload flow; per-dataset adapters provide only `download` + `parse` hooks. The first concrete adapter is SLURP/`s2s-v1`; `hf_source.py` adds a generic HF REST path (with a parquet fallback); `coval-build-dataset` is the CLI entry point.

- `balanced_sample` deduplicates, enforces per-dimension tag-coverage floors, excludes untagged clips, and does stratified round-robin; the tests cover all the documented edge cases.
- The metadata-first HF path pages only transcript/duration via the datasets-server API and downloads audio only for the selected clips, keeping bandwidth proportional to `num` rather than total dataset size.
- Path-traversal guards for `dataset_id` in both `_hf_spec` and `_write_manifest` prevent the manifest from being redirected outside the package directory.

The pipeline is well-structured and the critical paths (path-traversal validation, atomic file writes, chunked downloads, tag-floor abort) are all correctly implemented. No correctness bugs were found.

All findings are minor quality issues — a missing NoReturn annotation that forces a type: ignore suppression, a template-string that produces invalid Python when the dataset path contains a quote, and an unhandled OSError in the SLURP metadata downloader that bypasses the clean error-message path. None of these affect the build output or data integrity for normal inputs.

No files require special attention before merging; the scaffold_adapter template in hf_source.py and the download_slurp error handling in s2s_v1.py are the two spots worth a quick follow-up.

<sub>Reviews (3): Last reviewed commit: ["Harden parquet fallback and SLURP extrac..."](https://github.com/coval-ai/benchmarks/commit/5943f24c5dd3aef57321e133cc5c9cffbebe2926) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41517030)</sub>

### DIFF
--- a/runner/src/coval_bench/datasets/scripts/build.py
+++ b/runner/src/coval_bench/datasets/scripts/build.py
@@ -15,6 +15,8 @@ from __future__ import annotations
 
 import importlib.util
 import logging
+import re
+import shutil
 import sys
 from collections.abc import Callable
 from pathlib import Path
@@ -45,8 +47,18 @@ def _configure_logging() -> None:
 
 
 def _meta_dim(col: str) -> Callable[[Clip], object]:
-    """A balance-dimension accessor for meta column *col* (None when absent/empty)."""
-    return lambda clip: clip.meta.get(col) or None
+    """Balance-dimension accessor for meta column *col*.
+
+    Only genuinely missing/blank values count as untagged (``None``). ``False`` and
+    ``0`` are real values and are kept, so balancing on a boolean/numeric column
+    doesn't silently drop every ``False``/``0`` clip.
+    """
+
+    def dim(clip: Clip) -> object:
+        value = clip.meta.get(col)
+        return None if value is None or value == "" else value
+
+    return dim
 
 
 def _load_adapter_spec(name: str) -> DatasetSpec | None:
@@ -130,6 +142,8 @@ def _hf_spec(
     handwritten adapter only when neither can resolve it.
     """
     ds_id = dataset_id or f"{hf_path.split('/')[-1].lower()}-v1"
+    if not re.fullmatch(r"[A-Za-z0-9._-]+", ds_id):
+        raise click.UsageError(f"invalid dataset id {ds_id!r}: use letters, digits, '.', '_', '-'")
     download, parse, fetch = _resolve_hooks(
         hf_path, ds_id, config=config, split=split, audio_col=audio_col, text_col=text_col
     )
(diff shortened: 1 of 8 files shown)


runner/src/coval_bench/datasets/scripts/framework.py
```diff
@@ -0,0 +1,286 @@
...
+        "license": spec.license,
+        "source": spec.source,
+        "items": [
+            {
+                "path": f"audio/{c.filename}",
+                "sha256": c.sha256,
+                "transcript": c.transcript,
+                "duration_sec": round(c.duration_sec, 6),
+                "speech_end_offset_ms": None,
+                **{k: v for k, v in c.meta.items() if not k.startswith("_")},
```
coval-cale: If the meta data has any of the above fields, it'll overwrite that one.

runner/src/coval_bench/datasets/scripts/build.py
```diff
@@ -0,0 +1,229 @@
...
+            structlog.dev.ConsoleRenderer(),
+        ],
+        wrapper_class=structlog.make_filtering_bound_logger(logging.INFO),
+        logger_factory=structlog.PrintLoggerFactory(file=sys.stderr),
+    )
+
+
+def _meta_dim(col: str) -> Callable[[Clip], object]:
+    """A balance-dimension accessor for meta column *col* (None when absent/empty)."""
+    return lambda clip: clip.meta.get(col) or None
```
coval-cale: I think using `--balance` on a column with `False` or `0` value will just exclude all those values.

coval-cale (commented):
Two things I noticed, I feel like there might be some more but I figure it's best to get something out and then we can fix on the fly. Nice!

runner/src/coval_bench/datasets/scripts/framework.py
seribaymadina: nice catch, thanks for the review! im gonna close down all the unresolveds and from then on we will see

coval-cale (approved):
Sorry! Totally thought I approved this! I had a few nits but I think we should just use and change as needed

PR #227, opened 2026-07-07
[BENCH-369] Fix filter sidebar overlapping the footer on fast scroll

cooperr1:
Screenshot 2026-07-07 at 3 20 03 PM

Screenshot 2026-07-07 at 3 36 17 PM

## Problem
On the TTS/STT dashboards, scrolling quickly to the bottom made the filter chips (Deployment, Features, etc.) float into the footer. The sidebar was `position: fixed` and dodged the footer with a scroll-event-driven `translateY` — on fast/momentum scrolls the page moves on the compositor thread while the JS handler lags behind, so the correction arrives frames late (or never lands cleanly).

## Fix
Delete the scroll-sync effect entirely and put the sidebar in the content flow with `position: sticky`, so the footer pushes it up natively with zero JS:

- `ModelSidebar` becomes a plain sticky column inside the content row.
- `DashboardLayout` wraps sidebar + content in a flex row with a matching 18rem right spacer, keeping the content column centered exactly as before.
- Root `overflow-hidden` → `overflow-x-clip`, since `overflow-hidden` creates a scroll container that would break `sticky`; `clip` keeps the same horizontal clipping without one.

Net −53 lines.

### DIFF
--- a/web/components/layout/DashboardLayout.tsx
+++ b/web/components/layout/DashboardLayout.tsx
@@ -51,45 +51,48 @@ const DashboardLayout: React.FC<{ children: React.ReactNode }> = ({ children }) 
   }, [mode]);
 
   return (
-    <div className="relative flex min-h-screen flex-col overflow-hidden bg-background text-text-primary">
+    <div className="relative flex min-h-screen flex-col overflow-x-clip bg-background text-text-primary">
       <DashboardHeader />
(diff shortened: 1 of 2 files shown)


coval-cale (approved):
YEWWWWWW!!! love a good +/- ratio

PR #228, opened 2026-07-07
Add S2S fetch job, model registry entries, and chart colors

seribaymadina:
Stacks on #225 (the S2S benchmark axis + V2V metric) — merge that first.

Adds the daily Coval fetch-and-write job (`coval_bench/s2s/fetch_v2v.py`) that pulls each provider's newest completed run, reads the 50 per-clip V2V latency values, converts seconds to milliseconds once, and writes one result row per clip through the existing RunWriter so the matviews aggregate it exactly like STT/TTS. Registers the two realtime models (`openai/gpt-realtime`, `google/gemini-live`) in the model registry as pending so they stay hidden until the first run lands data, exposes the job as `coval-bench fetch-s2s` (fetch-only, so a standalone command rather than a `run --kind` value), adds the Coval API key/base to settings, and gives the two models chart colors.

Verified against the live Coval API: List filters by an AIP-160 string (`agent_id` + `status`), not query params, and tags aren't filterable. Nothing runs against Coval yet — blocked on the org, agents, and metric id, and there's no prod visibility (no `/s2s` route, models pending).

<h3>Greptile Summary</h3>

This PR adds the S2S Coval fetch path and registers the first realtime S2S models. The main changes are:

- New `fetch-s2s` CLI command for daily Coval V2V ingestion.
- Coval API settings for the fetch job.
- S2S model registry entries for OpenAI realtime and Gemini Live.
- Per-clip V2V fetch, result writing, and bucket refresh logic.
- Unit coverage for the new fetch flow.
- Chart colors for the new S2S models and Google provider.

This is close, but the retry repair path should be fixed before merging.

- A retry can skip `refresh_bucket()` when raw rows already exist.
- That can leave the daily S2S chart bucket empty or stale after a prior best-effort refresh failure.

runner/src/coval_bench/s2s/fetch_v2v.py

<sub>Reviews (10): Last reviewed commit: ["Annotate heterogeneous test fixtures for..."](https://github.com/coval-ai/benchmarks/commit/621a91872937ae544d367a4e9f651c2332077382) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42548594)</sub>

### DIFF
--- a/runner/src/coval_bench/s2s/fetch_v2v.py
+++ b/runner/src/coval_bench/s2s/fetch_v2v.py
@@ -1,0 +1,275 @@
+# Copyright 2026 The Coval Benchmarks Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Fetch S2S (voice-to-voice) latency from the Coval API and write per-clip rows.
+
+For each provider, find its newest completed Coval run, read the per-clip
+latency values (seconds), convert to milliseconds, and write one results row
+per clip via ``RunWriter``; the existing matviews aggregate. Agent ids, the
+metric id, and the API key are read from the environment, never committed.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import hashlib
+import importlib.resources
+from dataclasses import dataclass
+from datetime import UTC, datetime
+from typing import Any, cast
+
+import click
+import httpx
+import structlog
+
+from coval_bench.config import Settings, get_settings
+from coval_bench.db.conn import lifespan_pool
+from coval_bench.db.models import Result, ResultStatus, RunStatus
+from coval_bench.db.writer import RunWriter
+from coval_bench.registries import Metric
+from coval_bench.registries.benchmarks import Benchmark
+
+logger = structlog.get_logger("coval_bench.s2s.fetch_v2v")
+
+# Daily job: floor scheduled_at to the day so a run's clip rows share one bucket.
+DAILY_PERIOD_SECONDS = 86_400
+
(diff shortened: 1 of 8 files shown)


runner/src/coval_bench/s2s/fetch_v2v.py
```diff
@@ -0,0 +1,267 @@
...
+        )
+        if rows:
+            await writer.record_results(rows)
+        failed = sum(1 for r in rows if r.status is ResultStatus.FAILED)
+        if not rows:
+            status = RunStatus.FAILED
+        elif failed:
+            status = RunStatus.PARTIAL
+        else:
+            status = RunStatus.SUCCEEDED
```
coval-cale, line 209: Anything that fires after this line would mean that we're throwing away rows that actually succeeded.

runner/src/coval_bench/s2s/fetch_v2v.py
```diff
@@ -0,0 +1,267 @@
...
+
+async def fetch_and_write_v2v(settings: Settings | None = None) -> dict[str, RunStatus]:
+    """Fetch each provider's latest run into its own run row; return per-provider status.
+
+    A daily fetch replaces rather than accumulates: earlier runs in today's slot are
+    superseded first, so a retry or a re-pulled stale Coval run does not double-count.
+    """
+    settings = settings or get_settings()
+
+    metric_id = os.environ.get(METRIC_ID_ENV)
```
coval-cale: Maybe make this a .env variable instead of an os variable so that we're storing all our secrets in the same way.

coval-cale (approved):

PR #234, opened 2026-07-08
Show WER methodology note on the accuracy chart

coval-cale:
The whisper_normalizer methodology annotation from #231 never rendered: the latency timeline is the only surface that reads methodologyChanges, and it never displays WER, so a wer-tagged entry showed up nowhere. WER lives on the accuracy bar chart, which has no time axis.

This renders the note on the accuracy card (STT and TTS) while the selected data window spans the change, i.e. exactly while the bars aggregate non-comparable values, using the same gating as the timeline markers. The entry is also pinned to the real step: run 3908, 2026-07-08 01:00 UTC, the first run scored with the new normalizer.

Verified locally against the prod API on a production build: note renders on both dashboards, timeline markers unaffected by the shared-timestamp refactor.

<h3>Greptile Summary</h3>

This PR adds methodology-change notes to the accuracy chart. The main changes are:

- Adds a reusable `MethodologyNote` component for non-time-axis charts.
- Renders the WER methodology note in the accuracy bar section.
- Extracts methodology timestamp parsing into a shared helper.
- Pins the WER normalizer change to its explicit UTC-equivalent run time.

This looks safe to merge.

- No blocking issues found in the changed code.

<sub>Reviews (1): Last reviewed commit: ["Show WER methodology note on the accurac..."](https://github.com/coval-ai/benchmarks/commit/189118cc3aa72ce58b4206516d57e18ab5122910) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42586292)</sub>

### DIFF
--- a/web/components/dashboard/AccuracyBarSection.tsx
+++ b/web/components/dashboard/AccuracyBarSection.tsx
@@ -18,6 +18,7 @@ import CustomBarTooltip from "@/components/charts/tooltips/BarTooltip";
 import CustomBarChartTick from "@/components/charts/CustomBarChartTick";
 import Card from "@/components/shared/Card";
 import SectionHeader from "@/components/shared/SectionHeader";
+import MethodologyNote from "@/components/dashboard/MethodologyNote";
 import { useDashboard } from "@/contexts/DashboardContext";
 import { useThemeColors } from "@/hooks/useThemeColors";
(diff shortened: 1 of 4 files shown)


coval-cale:
Dropping this: the annotation approach for WER is not wanted.

PR #289, opened 2026-07-14
Enable Azure, Inworld, Alibaba, and MiniMax in the arena

seribaymadina:
Their keys are now mounted on benchmarks-api, so the arena_enabled opt-outs are stale. Extend check_arena_keys.py to fail on exactly that state, key mounted but provider still opted out,  so future flips aren't forgotten. Google stays out: ADC auth, no mountable key.

<h3>Greptile Summary</h3>

This PR expands the TTS arena roster and tightens the key parity check.

- Adds a reverse check for mounted keys with stale arena opt-outs.
- Enables several Azure, Inworld, Alibaba, Fish Audio, and MiniMax TTS models.
- Adds Azure to the provider API-key registry.

Azure arena battles can fail when the API key is mounted but the Azure region is not configured.

- The new roster path can select Azure TTS models.
- The Azure TTS provider requires `azure_region` during construction.
- The updated parity check only verifies provider API-key mounts, so it can miss this required runtime setting.

runner/src/coval_bench/registries/models.py

<sub>Reviews (1): Last reviewed commit: ["Enable Azure, Inworld, Alibaba, and Mini..."](https://github.com/coval-ai/benchmarks/commit/1349921c8ef890b7d7e40ca133169582416ceb29) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44257860)</sub>

### DIFF
--- a/runner/scripts/check_arena_keys.py
+++ b/runner/scripts/check_arena_keys.py
@@ -1,11 +1,15 @@
 # Copyright 2026 The Coval Benchmarks Authors
 # SPDX-License-Identifier: Apache-2.0
-"""Assert every arena-eligible TTS provider's API key is mounted on benchmarks-api.
+"""Assert arena roster and benchmarks-api key mounts agree, in both directions.
 
 Reads the ``coval-ai/benchmark-infra`` Cloud Run service module (passed as a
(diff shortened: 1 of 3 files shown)


coval-cale (approved):
LFG!

PR #301, opened 2026-07-15
Add a required domain dropdown to arena battle generation

seribaymadina:
Voters now pick a domain (Customer Service, Healthcare, Sales, Receptionist / Booking, Other) before generating a battle; the choice is stored on the battle and sticks across consecutive battles. The BFF validates the value against the same fixed list and forwards it to the runner.

Safe to deploy against the current backend, which already accepts an optional free-form domain — server-side validation lands separately. First of three: dropdown → domain validation/filter → backfill + require.

<h3>Greptile Summary</h3>

This PR requires voters to choose a domain when generating an arena battle. The main changes are:

- Adds a shared list of supported arena domains.
- Adds a required domain dropdown that persists across consecutive battles.
- Validates domains in the BFF and forwards them to the runner.
- Updates battle-source types, implementations, and unit tests.

<h3>Confidence Score: 5/5</h3>

This looks safe to merge.

- No blocking issues found in the changed code.
- The UI and BFF use the same domain values.
- The BFF rejects missing or invalid domains before calling the runner.

<sub>Reviews (1): Last reviewed commit: ["Add a required domain dropdown to arena ..."](https://github.com/coval-ai/benchmarks/commit/a57433774cde84630d3ab07f6c16137fcf4f13a0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44604146)</sub>

PR #303, opened 2026-07-15
Require a domain on new battles and backfill old ones to other

seribaymadina:
With the dropdown live, domain becomes required on battle creation. Migration 0011 tags the pre-dropdown battles as "other" — all 265 battles in prod have domain NULL today (verified against prod on 2026-07-15), and their 256 votes keep counting toward the aggregate "all" board, which never looks at domain. The snapshot CLI also gains --domain (default all), so the daily job can start computing per-domain boards.

Stacked on #302; retarget to main once it merges. Release the runner image for this one only after the dropdown (#301) is deployed on Vercel, otherwise battle generation from the old UI would 422. Last of three.

<h3>Greptile Summary</h3>

This PR requires domains for new battles and supports domain-specific leaderboard snapshots. The main changes are:

- Adds `--domain` to the snapshot CLI and its structured output.
- Requires a valid domain in battle creation requests.
- Backfills existing NULL battle domains to `other`.
- Adds coverage for missing-domain validation.

The API-backed web creation path will fail until it forwards the required domain.

- The runner now rejects the payload currently sent by the web proxy.
- The migration value matches the accepted domain set.
- Snapshot filtering and persistence receive the selected domain consistently.

runner/src/coval_bench/api/schemas.py and the web battle-creation proxy

<sub>Reviews (1): Last reviewed commit: ["Require a domain on new battles and back..."](https://github.com/coval-ai/benchmarks/commit/d88c6b0a4b5879990407ec64f492c8747f9e3c7f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44604617)</sub>

### DIFF
--- a/runner/src/coval_bench/__main__.py
+++ b/runner/src/coval_bench/__main__.py
@@ -330,12 +330,21 @@ def arena() -> None:
     help="RNG seed; keeps the bootstrap CI reproducible.",
 )
 @click.option(
+    "--domain",
+    default="all",
+    show_default=True,
+    type=click.Choice(
+        ("all", "customer-service", "healthcare", "sales", "receptionist-booking", "other")
+    ),
+    help="Domain board to compute ('all' aggregates every battle regardless of domain).",
+)
+@click.option(
     "--force",
     is_flag=True,
     default=False,
     help="Run even if another snapshot is in progress (bypass the advisory lock).",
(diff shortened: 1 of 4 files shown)


coval-cale (approved):
Epic!

PR #306, opened 2026-07-15
Split TTS run samples across a female and a male voice

seribaymadina:
TTS runs previously synthesized all 10 samples with one pinned voice per model, so every metric reflected a single (usually US-female) voice. Each active TTS model now carries a `voices` pool ordered (female, male), and the orchestrator splits the run's samples evenly across it  exactly 5/5, with which transcript lands on which voice shuffled per run using a `run_id`-derived seed, so the pairing rotates across runs but stays reproducible within one. The assigned voice is recorded on every result row (the `voice` column already existed), so per-gender analysis needs no schema change and no migration.

A methodology marker dated 2026-07-15 17:00 PT flags the flip on the timeline; the time should be corrected to the actual runner-release time at merge, since single-voice and split-voice values are not directly comparable.

Additional: Some models change voice with this PR because the old pin didn't fit a US female/male pair: ElevenLabs (Charlie is Australian, and legacy premade voices are being replaced by Dec 31), Hume (the pin turns out to be a British male, "Nature Documentary Narrator"), OpenAI, Soniox, MiniMax, and Gradium (the pin was a docs-example ID absent from their catalog). Deepgram keeps its single voice its voice IS the model string, so a second voice means a second registry entry, deferred.

<h3>Greptile Summary</h3>

This PR splits TTS benchmark samples between female and male voices. The main changes are:

- Adds balanced voice pools to active TTS models.
- Assigns voices deterministically for each run.
- Records the selected voice on every result row.
- Keeps TTS tasks in item-major order across providers.
- Adds a timeline marker for the methodology change.

The timestamp should be corrected before merging.

- The provider request ordering has been corrected.
- The methodology marker still uses a provisional time and can place runs on the wrong side of the change boundary.

web/lib/config/methodologyChanges.ts

<sub>Reviews (2): Last reviewed commit: ["Keep TTS task order item-major so provid..."](https://github.com/coval-ai/benchmarks/commit/db3f16b076234e5f7c6b4813ede0ef732487c4b8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44645728)</sub>


coval-cale (approved):
Epic!

PR #310, opened 2026-07-16
Add portable benchmark run artifacts

RitwijParmar:
This adds an opt-in JSONL artifact for benchmark runs.

The DB stays the source of truth. The artifact is for debugging one run after the fact, especially when a provider fails or a run finishes partial.

It writes:
- one run header with runner and dataset metadata
- one sanitized record per metric row
- one summary record with success and failure counts plus grouped failure buckets

I kept transcripts out of the artifact. It stores only sha256 plus character count. That gives enough identity for debugging without dumping raw transcript text into local files.

Enable it with `RUN_ARTIFACT_DIR`. If unset, behavior is unchanged.

Review updates pushed in `d68e385`:
- the Docker example now bind-mounts the artifact directory so output survives `--rm`
- transcript hashing now ignores non-string values safely
- artifact writes run off the event loop
- failed-run DB finalization happens before best-effort artifact output

Checked locally:
- `uv run pytest tests/unit/test_run_artifacts.py tests/unit/test_orchestrator.py`
- `uv run ruff check src/coval_bench/config.py src/coval_bench/runner/artifacts.py src/coval_bench/runner/orchestrator.py tests/unit/test_run_artifacts.py tests/unit/test_orchestrator.py`
- `uv run ruff format --check src/coval_bench/config.py src/coval_bench/runner/artifacts.py src/coval_bench/runner/orchestrator.py tests/unit/test_run_artifacts.py tests/unit/test_orchestrator.py`
- `uv run mypy src/coval_bench/runner/artifacts.py src/coval_bench/runner/orchestrator.py`

### DIFF
--- a/runner/README.md
+++ b/runner/README.md
@@ -46,6 +46,10 @@ docker compose up -d api         # FastAPI on http://localhost:8000
 # Trigger a single-item benchmark run (writes to the local Postgres):
 docker compose run --rm runner coval-bench run --smoke --kind tts
 
+# Optional: write a portable JSONL artifact for the run:
+docker compose run --rm -e RUN_ARTIFACT_DIR=/tmp/artifacts runner \
+  coval-bench run --smoke --kind tts
+
 # Probe one TTS provider without DB writes:
 docker compose run --rm runner coval-bench tts-smoke \
   --provider cartesia --model sonic-3 --voice <voice-id> --text "hello"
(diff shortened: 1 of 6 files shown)


RitwijParmar:
The artifact is useful because a failed voice benchmark usually gets debugged after the run is gone. The next piece I would add is a replay command that takes one JSONL artifact and re-runs only the failed cases with the same provider settings and artifact schema. That would let someone compare a provider failure against a later run without touching the whole dataset. I would keep it opt-in and make the artifact self-describing enough to reject a mismatched runner or dataset.

callumreid (approved):

PR #365, opened 2026-07-23
Hide dedicated-inference models from dashboards

cooperr1:
## What changed

- Exclude models tagged `dedicated-inference` when the frontend builds dashboard model lists.
- Apply the exclusion to both catalogue-backed models and data-backed aggregate entries.
- Add a regression test covering Baseten Whisper STT and Qwen TTS alongside
  shared Deepgram and OpenAI models.

## Why

Baseten traffic was exercised through a non-persisting probe, so there is no persisted benchmark run backing public TTFA/WER values. Dedicated endpoints should stay off the public dashboards until the benchmark persistence and publication flow is agreed and implemented.

This is intentionally frontend-only. It does not change the API, runner,
registry, or provider integrations. The frontend contains no synthetic or
hardcoded benchmark values; dedicated STT and TTS aggregate rows are ignored
even if they are present.

## Impact

Dedicated-inference models no longer appear in dashboard filters, charts, or comparison tables. Shared models are unaffected.

## Validation

- `pnpm test lib/utils/modelsFromResults.test.ts` — 9 passed
- `pnpm typecheck`
- `pnpm exec eslint lib/utils/modelsFromResults.ts lib/utils/modelsFromResults.test.ts --max-warnings 0`

### DIFF
--- a/web/lib/utils/modelsFromResults.test.ts
+++ b/web/lib/utils/modelsFromResults.test.ts
@@ -105,6 +105,47 @@ describe("buildModelsByProvider", () => {
     expect(out.openai).toEqual(["openai:gpt-realtime-whisper"]);
   });
 
+  it("does not include dedicated-inference catalogue models or data-backed entries", () => {
+    const catalogue = {
+      tts: [
+        {
+          provider: "baseten",
+          models: [
+            {
+              model: "qwen3-tts-1.7b",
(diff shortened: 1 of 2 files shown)


cooperr1:
Superseded by #366, which points at the verified frontend-only STT/TTS dedicated-model guard.

PR #400, opened 2026-07-27
Screen arena prompts before synthesis

seribaymadina:
Reject payment cards and national ids locally, then run the prompt through OpenAI moderation. Both checks run ahead of the daily cap so a rejected prompt costs neither credits nor anyone's quota.

Moderation fails open, which leaves the local check as the backstop. It also does not treat scam framing as harmful, so the two cover different ground and both are needed.

The PII match is narrow on purpose: a blanket digit-run rule rejected one of the hundred curated example prompts, so cards are Luhn-checked and phone numbers and order refs are allowed.

<h3>Greptile Summary</h3>

This PR adds local card/national-ID detection and OpenAI prompt moderation before arena quota checks and synthesis.
- Adds configurable fail-closed handling for unavailable moderation.
- Extracts the arena domain literal into the arena package.
- Adds API and unit coverage for rejection, outage, parsing, timeout, and credential-rotation behavior.

The PR should not merge until credential rotation can no longer close a moderation client that an active request is using.

The parsing failure and stale-key behavior from the previous threads are addressed, but the replacement logic now closes a process-global client without coordinating with concurrent moderation calls, so requests overlapping key rotation can be misclassified as moderation outages.

**Files Needing Attention:** runner/src/coval_bench/arena/moderation.py

<sub>Reviews (3): Last reviewed commit: ["Close the superseded moderation client a..."](https://github.com/coval-ai/benchmarks/commit/997f422966f407848de21da4b9adcd8ac361e202) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47775550)</sub>

**Context used:**

- Knowledge Base — [API service](https://app.greptile.com/coval/-/custom-context/knowledge-base/coval-ai/benchmarks/-/docs/api-service.md)
- Knowledge Base — [Arena subsystem](https://app.greptile.com/coval/-/custom-context/knowledge-base/coval-ai/benchmarks/-/docs/arena-subsystem.md)

### DIFF
--- a/runner/src/coval_bench/api/routers/arena.py
+++ b/runner/src/coval_bench/api/routers/arena.py
@@ -40,7 +40,6 @@ from starlette.requests import Request
 from coval_bench.api.deps import capture_api_event, get_pool, get_posthog, get_settings
 from coval_bench.api.ratelimit import limiter
 from coval_bench.api.schemas import (
-    ArenaDomain,
     ArenaLeaderboardResponse,
     BattleCreate,
     BattleOut,
@@ -52,6 +51,7 @@ from coval_bench.api.schemas import (
     VoteOut,
 )
 from coval_bench.arena.audio_store import clip_url
+from coval_bench.arena.domains import ArenaDomain
 from coval_bench.arena.generate import generate_battle
 from coval_bench.arena.moderation import moderation_verdict, pii_match
 from coval_bench.arena.monitoring import (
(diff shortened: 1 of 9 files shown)


callumreid (approved):

seribaymadina:
Pushed c6228b58.

| Finding | Fix |
|---|---|
| P1 timeout | `timeout=4.0, max_retries=0`, plus our own `asyncio.timeout` so pool/DNS stalls are bounded too, not just the HTTP read. Test hangs a stub for 30s. |
| P1 CI | Root fix, not suppressed: `ArenaDomain` → new `arena.domains`, so `arena.prompts` stops importing the FastAPI app and cycling through `api.routers.arena`. `mypy --strict src tests` clean over 232 files. |
| P2 fail-open | Switched to fail **closed** — 503 when unreachable. The local check finds card and id numbers, not harmful content, so it was never a content-safety fallback. `arena_moderation_fail_closed` buys availability back. |
| Greptile: parsing outside the guard | Fixed — parsing moved inside the `try`, so a malformed reply is an unavailable moderator, not a 500. |
| Greptile: client keeps first key | Fixed — cached against a SHA-256 of the key, rebuilt when it changes. |
| Luhn false positives | Cards now need a known issuer prefix as well as the checksum, so a 16-digit tracking reference no longer reads as a card. |

Deliberately not here: **BFF error mapping** is a web change and ships with the next web PR — the runner now returns 503 moderation-unavailable, distinct from 502 synthesis and 429 quota, so the statuses to map exist. **Attempts table, per-person caps and the kill switch** need a migration and are the next PR.

seribaymadina:
@coderabbitai review

PR #403, opened 2026-07-28
Mint a signed identity for every arena visitor

seribaymadina:
The arena opened to the public in #395, and public visitors get no cookie at all: `isPublicArenaPath` returns a bare `NextResponse.next()`, so the gate that used to mint one never runs. Votes are therefore attributed to a UUID the browser generates in localStorage, which means a vote cannot be tied to a person, held to a quota, or judged by later scoring.

This mints a signed identity for everyone instead. The cookie payload gains a server-generated `sid` and a `role` of `labeler` or `external`, both covered by the existing HMAC so neither can be forged. Refreshing preserves them minting a fresh `sid` every 24 hours would weaken vote dedupe and strip any cross-day signal from scoring.

Cookies predating identity belong to the internal labelers who were using the arena before it opened, so they are reissued as `labeler`; everyone arriving since is `external`. Nobody is logged out and no link needs resending.

**Nothing reads `sid` or `role` yet.** The BFF still takes `voterId` from the request body, so voting and the board are unchanged by this PR

### DIFF
--- a/web/.env.example
+++ b/web/.env.example
@@ -38,6 +38,11 @@ ARENA_ACCESS_TOKEN=
 # HMAC key for the access cookie. 32+ random bytes, base64. Generate with:
 #   node -e "console.log(require('crypto').randomBytes(32).toString('base64'))"
 ARENA_SESSION_SECRET=
+# The outgoing ARENA_SESSION_SECRET, accepted for verification only. Rotating without it
+# invalidates every live cookie at once: labelers lose access and every voter is issued a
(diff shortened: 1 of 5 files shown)


cooperr1:
Two things before merge: setIdentityCookie now goes through buildCookieHeader instead of res.cookies.set, which set secure: true and sameSite: "lax" explicitly, so those should still get appended. Also, ensureArenaIdentity grants labeler to any sid-less cookie on an unauthenticated path, which is only safe because legacy cookies required the unlock token. Could use an explicit migration with a cutoff.

cooperr1 (approved):

PR #405, opened 2026-07-28
Lay the box plot's caption outside the scroller (BENCH-552)

cooperr1:
Fixes [BENCH-552](https://linear.app/coval/issue/BENCH-552/ranked-by-median-latency-caption-lagsre-renders-during-horizontal).

## The bug

In Latency Variation, the "Ranked by median latency (fastest first)" caption trailed the horizontal scroller and snapped into place once scrolling stopped.

It was an SVG `<text>` inside the scrolling plot, counter-translated by a `scrollX` React state set in `onScroll` and applied in a `useLayoutEffect`. Confirmed in the browser: setting `scrollLeft = 300` left the node at `translate(0, 474)`, and it only reached `translate(300, 474)` after the next commit. That one-commit delay is the drift-then-snap.

## The fix

The caption is chart chrome, not plot content, so it no longer lives in the scroller at all — it's a mono `<p>` absolutely positioned over the bottom band `margin.bottom` already reserves, in the non-scrolling container. Zero scroll coupling, so there is nothing left to lag: no scroll handler, no rAF, no transform bookkeeping.

The PNG export keeps the caption via `exportXLabel` on `SectionHeader`, the same mechanism the WER bars already use (`QualityBarSection.tsx:184`). That also fixes the old export, which captured the caption at whatever scroll offset the reader happened to leave it at.

Net −37/+16: the scroll-tracking `useLayoutEffect`, `captionRef`, `captionY`, and `axisLabelFontSize` are all gone.

## Verification

At 375 / 1280 / 1440 / 2400 px, on `/tts` and `/stt` (22 models — worst overflow), scrolled to both extremes:

- Caption's viewport x is identical before and after a scroll (`dx: 0`) at every width.
- `capBottom === frameBottom` everywhere — no layout growth, card height unchanged.
- Centered when the plot fits, left-anchored at the plot edge when it scrolls (prior behavior preserved); short form on mobile.
- Geist Mono 14px, `rgb(219,219,211)` — styling identical to the old SVG text.
- `pointer-events-none`, so tap-to-pin on a box still works on mobile.
- `tsc --noEmit` and `eslint` clean.

The `median latency` vs `IQR` copy mismatch the ticket mentions is left alone — tracked in [BENCH-550](https://linear.app/coval/issue/BENCH-550/ttfa-distribution-chart-sorted-by-legacy-quartile-metric-not-iqr).

### DIFF
--- a/web/components/charts/d3/BoxPlot.tsx
+++ b/web/components/charts/d3/BoxPlot.tsx
@@ -19,9 +19,12 @@ const modelFontSize = 12;
 const providerFontSize = 12;
 const yAxisTickFontSize = "12px";
 const modelLineHeight = 14;
-// Bottom holds up to three label lines, the provider, the dedicated-inference
-// marker, and the axis caption — 80px stacked marker over caption.
-const margin = { top: 20, right: 8, bottom: 88, left: 40 };
+// Bottom holds the whole label stack: 15px clear of the plot, the label lines
+// (four of them for a name like Voxtral-Mini-Transcribe-2602), the provider,
+// then the dedicated-inference marker, 15px tall and 4px under the provider.
+// The caption is not in here — it gets the container's own bottom padding, so
+// the deepest stack can never land on it however the labels wrap.
+const margin = { top: 20, right: 8, bottom: 100, left: 40 };
 const minSlotWidth = 48;
 /** Share of a slot the label block may occupy; the rest is breathing room. */
 const labelBandRatio = 0.82;
@@ -503,7 +506,7 @@ const BoxPlot: React.FC<BoxPlotProps> = ({
   // leave the sizing effect's ResizeObserver attached to nothing, freezing
   // dimensions at their initial value once data arrives.
   return (
-    <div ref={containerRef} className="relative w-full" data-export-frame>
+    <div ref={containerRef} className="relative w-full pb-5" data-export-frame>
       {data.data.length === 0 ? (
         // Same height as the populated chart so toggling to a model with no
         // latency runs doesn't shift the sections below.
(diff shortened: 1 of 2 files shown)


cooperr1:
Pushed `74b710b` — the caption could still be overlapped on mobile, which the first commit didn't address.

**What was wrong:** the caption was overlaid on the bottom band `margin.bottom` reserves for axis labels, and that band was undersized for the deepest stack. A model name can wrap to **four** lines (`Voxtral / Mini / Transcribe / 2602`), and a dedicated-inference marker hangs 15px more below the provider. Measured on `/stt` at 375px, the deepest provider label ran **7.9px into** the caption's line box, and a marker under it would have spilled past the plot's bottom edge entirely.

**Fix:** the caption now gets the container's own bottom padding (`pb-5`) rather than a slice of the plot's margin, and `margin.bottom` is sized for the label stack alone (88 → 100, which also keeps the marker inside the plot box). Nothing inside the SVG can reach the caption now, however the labels wrap. The padding sits on the container, so the empty state reserves the same strip and the sections below still don't shift.

Re-measured at 320 / 375 / 1280 / 2400 px on `/tts` and `/stt` (1d and 30d), scrolled to both extremes:

- `capTop === svgBottom` — the caption starts exactly where the plot ends, so overlap is structurally impossible.
- Deepest label clearance **+24.1px** (was −7.9px); a marker simulated under that same four-line label clears by **+7.3px**.
- `dxOnScroll: 0` still holds — the original lag fix is unaffected.
- Caption fits on one line at 320px (206px of 280px available), no clipping.
- `tsc --noEmit` and `eslint` clean.

coval-cale (approved):
YEWWWWWW

PR #409, opened 2026-07-28
[BENCH-557] Read arena key parity from the infra provider map

coval-cale:
The gate now fetches envs/prod/provider_keys.tf and reads api = true entries, falling back to the service module's inline env blocks for older infra revisions.


seribaymadina (approved):
noiceee

PR #414, opened 2026-07-29
Scope the early-access embargo to the caller

seribaymadina:
The embargo was all-or-nothing: one key revealed every unreleased model. A request can now carry a scope naming the providers it may see, honoured only alongside its own key, so a partner sees their own numbers and no one else's. An absent scope behaves exactly as before.

Note: Partner gated, tempo solution until we have proper auth set up


coval-cale (approved):

runner/src/coval_bench/api/internal.py
```diff
@@ -38,3 +62,32 @@ def is_internal(
...
+    honoured only alongside its own key, which is separate from the internal one
+    so a narrowed caller can never reach unrestricted access. Names outside the
+    embargoed set are dropped rather than rejected, so a malformed header is
+    indistinguishable from an absent one.
+    """
+    if internal:
+        return UNRESTRICTED
+    if not _matches(x_ea_key, settings.early_access_bff_key):
+        return frozenset()
+    return frozenset((x_ea_scope or "").split(",")) & embargoed_providers()
```
cooperr1: (x_ea_scope or "").split(",") doesn't strip whitespace, so "a, b" silently drops b. 

The API's access rule is "present the BFF key, name any embargoed provider you want." Partner isolation lives entirely in the BFF; a leaked or misrouted key grants the entire embargo, and the API can't tell partners apart, so there's also no per-partner audit trail.

runner/tests/api/test_early_access_scope.py
```diff
@@ -0,0 +1,179 @@
...
+    """The public view: an anonymous caller sees none of them."""
+    assert hidden_models_for(frozenset()) == _embargoed_pairs()
+    assert hidden_models() == _embargoed_pairs()
+
+
+def test_unrestricted_scope_hides_nothing() -> None:
+    assert hidden_models_for(UNRESTRICTED) == frozenset()
+
+
+@pytest.mark.parametrize("provider", sorted(embargoed_providers()))
```
cooperr1: evaluates the registry at collection time. fine today, just less explicit than the _one_embargoed_provider() skip pattern used elsewhere.

cooperr1 (approved):

PR #417, opened 2026-07-30
[BENCH-569] Simplify arena leaderboard CSV export and improve table formatting

cooperr1:
## Summary

Example output: 
Rank,Provider,Model,Rating,Rating Margin (±),Votes,Wins,Losses,Ties,Status
1,Cartesia,Sonic 3.5,1709,183,21,15,3,3,preliminary
2,ElevenLabs,Flash v2.5,1566,138,21,11,7,3,preliminary
3,OpenAI,GPT-4o mini TTS,1434,129,21,7,11,3,preliminary
4,Deepgram,Aura 2,1291,182,21,3,15,3,preliminary

- CSV export now uses friendly, human-readable columns: `Rank, Provider, Model, Rating, Rating Margin (±), Votes, Wins, Losses, Ties, Status` — ratings rounded to whole numbers, provider/model as display names, `rating_bt` and the three raw CI columns dropped in favor of a single margin column
- CSV generation reuses the shared `downloadCSV`/`serializeCSV` helpers (BOM for Excel, deferred blob revoke) instead of the component's hand-rolled serializer
- The table's ± value gets the shared `MetricInfo` tooltip explaining the confidence interval — works on tap (mobile), hover, and keyboard focus, with the panel opening above the trigger
- The mixed-status "provisional" marker is now a small badge instead of plain inline text

## Testing
- `pnpm test` (112 passed), `pnpm lint`, `pnpm typecheck` all clean
- Verified against the local full-stack arena: CSV blob content matches the new schema exactly; tooltip opens on tap at mobile width and dismisses on outside tap

### DIFF
--- a/web/app/overview/arena/leaderboard/arena-leaderboard.tsx
+++ b/web/app/overview/arena/leaderboard/arena-leaderboard.tsx
@@ -9,6 +9,8 @@ import DashboardHeader from "@/components/layout/DashboardHeader";
 import DashboardFooter from "@/components/dashboard/DashboardFooter";
 import Card from "@/components/shared/Card";
 import { CymaticLoader } from "@/components/shared/CymaticLoader";
+import MetricInfo from "@/components/shared/MetricInfo";
+import { downloadCSV } from "@/lib/utils/chartExport";
 import {
   normalizeModelName,
   normalizeTTSProviderName,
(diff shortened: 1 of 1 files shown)


seribaymadina (approved):
nice! do you see any value in adding /overview/leaderboard endpoint?

PR #420, opened 2026-07-30
[BENCH-571] Hide xAI from S2S and launch S2S to live

cooperr1:
## What

Launches the S2S benchmark publicly and hides xAI from it.
Screenshot 2026-07-30 at 1 08 02 PM

**Runner registry**
- `gpt-realtime` (OpenAI) and `gemini-live` (Google) leave the `EARLY_ACCESS` pre-launch embargo → `ACTIVE`.
- Both xAI models (`grok-realtime`, `grok-voice-think-fast-2.0`) → `PENDING`: hidden sitewide while xAI is unresponsive to outreach; the fetch job doesn't read the registry, so data keeps accruing for the flip back (same pattern as #322).

**Web launch surfaces**
- Removed the `NEXT_PUBLIC_S2S_ENABLED` gate from `/s2s` (the deployment env var is now dead config and can be cleaned up separately).
- Added Speech-to-Speech to the header tabs, mobile menu, footer, and overview hero copy.
- New third overview leaderboard card ranking V2V p50, pinned to the `s2s-multiturn-v1` dataset (hoisted to a named constant shared with the dashboard) and clamped to 7d when the shared toggle sits on 1d — S2S ingests daily and has no 24h window.

**Bug fix caught along the way**
- Data endpoints only strip `EARLY_ACCESS` models; `PENDING`/`RETIRED` rows still come back in `model_stats`. The overview cards ranked raw stats, so pending xAI (which keeps ingesting) would have surfaced there. Added `disabledModelKeys` and excluded disabled catalogue models from all three overview cards — closing the same latent gap for TTS/STT.

## Testing

- 432 runner unit tests, 113 web tests, typecheck + lint all green.
- Verified against a local runner API running this registry: `/v1/providers` serves xAI as `disabled`, OpenAI/Google enabled; `/s2s`, nav, footer, and overview card all render; overview card requests/labels 7d on the 1d toggle.
- Verified with prod data (providers catalogue from local API, data endpoints from prod): all S2S surfaces — overview card, key metrics, charts, comparison table, conversation samples — show OpenAI + Google only, no Grok.

## Rollout

The registry flip takes effect when the runner API redeploys; the web changes are independent and safe to deploy alongside.

### DIFF
--- a/runner/src/coval_bench/registries/models.py
+++ b/runner/src/coval_bench/registries/models.py
@@ -798,36 +798,37 @@ MODEL_REGISTRY: list[RegisteredModel] = [
     # S2S #
     #######
     # S2S realtime models. Numbers are fetched daily from Coval (no local
-    # provider client). EARLY_ACCESS = pre-launch embargo: they run and fetch
-    # normally, but every data endpoint strips them for public callers and
-    # serves them only to X-Internal-Key requests until the benchmark launches.
+    # provider client).
     RegisteredModel(
         benchmark=_S2S,
         provider="openai",
         model="gpt-realtime",
         tags=(_STREAMING, _MULTI),
-        status=_EARLY_ACCESS,
+        status=_ACTIVE,
     ),
     RegisteredModel(
         benchmark=_S2S,
         provider="google",
         model="gemini-live",
         tags=(_STREAMING, _MULTI),
-        status=_EARLY_ACCESS,
+        status=_ACTIVE,
     ),
+    # xAI hidden for launch while they are unresponsive to outreach; the fetch
+    # job doesn't read this registry, so data keeps accruing for the flip back
+    # to active.
     RegisteredModel(
         benchmark=_S2S,
         provider="xai",
         model="grok-realtime",
         tags=(_STREAMING, _MULTI),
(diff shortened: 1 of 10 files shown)


runner/src/coval_bench/registries/models.py
```diff
@@ -807,36 +798,37 @@ class RegisteredModel(BaseModel, frozen=True, extra="forbid"):
...
+    # xAI hidden for launch while they are unresponsive to outreach; the fetch
+    # job doesn't read this registry, so data keeps accruing for the flip back
+    # to active.
     RegisteredModel(
         benchmark=_S2S,
         provider="xai",
         model="grok-realtime",
         tags=(_STREAMING, _MULTI),
-        status=_EARLY_ACCESS,
+        status=_PENDING,
```
seribaymadina: PENDING doesn't seem to hide the xAI data it can be retrieved, lets keep EARLY_ACCESS

seribaymadina: note: im currently working on making gcs bucket private, so dont worry about that part

seribaymadina (approved):
approved, but please change the status of xAI models

PR #422, opened 2026-07-30
Resolve early access from a Clerk session token

coval-cale:
A signed-in provider org's benchmark_providers claim names the providers whose embargoed models it may see; anything unverified keeps the public view, and nothing activates until clerk_issuer is set.

Prep for adding early access orgs.

### DIFF
--- a/runner/src/coval_bench/api/internal.py
+++ b/runner/src/coval_bench/api/internal.py
@@ -36,9 +36,9 @@ logger = structlog.get_logger("coval_bench.api.internal")
 # Which proof the response honoured: internal, accepted, unknown, or absent.
 EA_STATUS_HEADER = "X-EA-Token-Status"
 
-# Both proof headers. Listing one is worse than listing none: a cached internal
-# response carries no X-EA-Token, so it would match a public request.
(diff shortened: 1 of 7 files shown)


runner/src/coval_bench/api/internal.py
```diff
@@ -136,6 +139,14 @@ def hidden_early_access(
         return frozenset()
 
     embargoed = embargoed_pairs()
+    if x_ea_token is None and authorization is not None and settings.clerk_issuer is not None:
```
seribaymadina, line 142: can we check if any possible cache issues

runner/src/coval_bench/api/internal.py
```diff
@@ -122,6 +124,7 @@ def hidden_early_access(
     response: Response,
     internal: bool = Depends(is_internal),
     x_ea_token: str | None = Header(default=None),
+    authorization: str | None = Header(default=None),
```
seribaymadina, line 127: yeah lets js be mindful of where the frontend obtain the Clerk session token

seribaymadina (approved):
good job! left comments to keep in mind/questions

PR #424, opened 2026-07-30
[BENCH-582] Track S2S sample engagement in PostHog

cooperr1:
## Summary

Adds PostHog tracking for the new speech-to-speech section's unique surface — the conversation-samples player. The shared dashboard chrome (time window, facets, chart hover/share, scroll depth, heatmap sort) already reports with `surface: "s2s_dashboard"` for free, so this PR covers what's S2S-specific:

- **`s2s_sample_play_requested` enriched** with `trigger` (`button` | `timeline`), `model_id`, `position`, and `has_transcript`. Previously a timeline-tooltip autoplay and a deliberate Play press were indistinguishable.
- **`s2s_sample_playback_ended`** (new): one listen session per activation, reporting `listen_pct` (furthest playhead), `duration_seconds`, `completed`, and the starting `trigger`. Flushed on pause, stop, track swap, end, and unmount — answers "did they actually listen or bail?"
- **`s2s_sample_seeked`** (new): transcript-turn clicks (`method: "turn"`, with `turn_index`/`turn_role`) and scrub-slider gestures (`method: "slider"`, debounced to one event per burst at its final position).
- **`s2s_sample_tick_changed`** (new): the older/newer chevrons and "Return to latest", with `direction` and `from_bucket`/`to_bucket`.
- **`dashboard_wer_bar_clicked` tagged with `metric`** (`wer` | `instruction`). Note: S2S instruction bars aren't clickable today, so only `"wer"` can fire — the tag keeps the shared section honest if they ever become clickable.

Attribution is snapshot-based: `playback_ended` and deferred slider seeks capture bucket/model at interaction time, so flushes that outlive a tick swap attribute to what was actually heard, not the manifest that replaced it.

## Test plan

- `pnpm typecheck`, `pnpm lint`, `pnpm test` (124 tests) all green.
- Verified live against the dev server with a `posthog.capture` spy (dummy token, nothing sent to the real project):
  - Play → `trigger: "button"`, `model_id`, `position`, `has_transcript` correct
  - Pause at ~18s of 80s → `listen_pct: 22`, `completed: false`
  - Seek near end, play out → `listen_pct: 100`, `completed: true`
  - Transcript turn click → `method: "turn"`, `turn_index: 1`, `turn_role: "agent"`
  - Slider drag → single debounced event at final position
  - Chevron/latest navigation → `direction: "older"` / `"latest"` with correct buckets
  - Scrub then tick-swap within the debounce window → seek still attributed to the scrubbed manifest
  - TTS WER bar click → `metric: "wer"`
- Not exercised live: `trigger: "timeline"` (recharts tooltip resists automation) — it's a one-line ref set inside the pre-existing autoplay effect.

### DIFF
--- a/web/app/s2s/components/ConversationTurns.tsx
+++ b/web/app/s2s/components/ConversationTurns.tsx
@@ -32,7 +32,7 @@ export function ConversationTurns({
   turns: S2STurn[];
   accentColor: string;
   currentTime?: number;
-  onSeek?: (seconds: number) => void;
+  onSeek?: (seconds: number, turn: { index: number; role: "caller" | "agent" }) => void;
 }) {
   const active = onSeek ? activeTurnIndex(turns, currentTime) : -1;
   // Only fade the others once something is genuinely highlighted. Without
@@ -68,7 +68,12 @@ export function ConversationTurns({
           <button
             key={t.index}
             type="button"
-            onClick={() => onSeek(t.start_offset as number)}
+            onClick={() =>
+              onSeek(t.start_offset as number, {
+                index: t.index,
+                role: agent ? "agent" : "caller",
+              })
+            }
             aria-current={isActive ? "true" : undefined}
             className={`${className} block w-full cursor-pointer text-left hover:opacity-100`}
             style={style}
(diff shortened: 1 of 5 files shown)


coval-cale (approved):
LFG!

PR #441, opened 2026-08-04
Alert when a run fails outright or a provider fails everything it runs

seribaymadina:
The orchestrator stored a run's status without ever emitting the `RUN_FAILED` / `RUN_PARTIAL` strings the alert metrics match on, so a run that completed and produced nothing paged nobody. Eight `ACTIVE` models across five providers have been failing every item since 29 July, unnoticed.

A total failure now reports `RUN_FAILED`. A partial run reports `RUN_PARTIAL` only when some provider failed **every one of its items**, named with its cause  `TTS:hume/octave-2 (server rejected WebSocket connection: HTTP 429)`. That threshold is deliberate: gladia and assemblyai each lost one clip in three days, and paging on those would get the alert muted. Neither path raises, since `max_retries = 1` means a non-zero exit re-runs the whole benchmark and re-spends on every provider.

Reported after `finish_run`, otherwise a failed write pages twice for one incident. The SIGTERM path reports too, limited to providers from a phase that finished before the cancellation.

Needs coval-ai/benchmark-infra#87 applied first so the events have a metric to land in.

### DIFF
--- a/runner/src/coval_bench/runner/orchestrator.py
+++ b/runner/src/coval_bench/runner/orchestrator.py
@@ -41,6 +41,7 @@ import importlib
 import random
 import signal
 import wave
+from collections import Counter, defaultdict
 from datetime import UTC, datetime  # noqa: UP017 — UTC alias requires 3.11+, target is 3.12
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal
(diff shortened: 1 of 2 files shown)


coval-cale (approved):
Sick but super dense comments. Let's trim them down a bit. They've got that high-volume low-density thing that LLMs do that make it tough to understand what's actually going on.

PR #58, opened 2026-06-03
Fix aiohttp Dependabot alert

kobicovaldev:
## Summary
- Bump runner aiohttp dependency floor to 3.14.0 to address GHSA-jg22-mg44-37j8 / CVE-2026-34993
- Regenerate uv.lock so aiohttp resolves to 3.14.0

## Validation
- uv lock --check
- uv run pytest (425 passed, 1 skipped)

## Summary by CodeRabbit

* **Chores**
  * Updated dependency to a more recent stable version to improve performance and stability.

PR #38, opened 2026-05-28
Update .gitignore so we can keep local dev scripts

coval-cale:
## Summary by CodeRabbit

* **Chores**
  * Updated ignore rules to exclude a local directory used for temporary/generated files, reducing accidental commits and repository clutter.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coval-ai/benchmarks/pull/38?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #448, opened 2026-08-05
chole(deps): bump aiohttp to 3.14.3

borgesius:
Clears alerts #63, #64 and #65 — the same aiohttp advisory trio patched in [backend#6437](https://github.com/coval-ai/backend/pull/6437), which surfaced here too.

| Advisory | Severity |
|---|---|
| Out-of-bounds heap read in the C HTTP response parser error path (malformed chunked response) | high |
| WebSocket client accepts compressed frames without negotiated permessage-deflate | moderate |
| HTTP request smuggling via WebSocket upgrade | moderate |

3.14.3 is the first release patching all three.

## Scope

`runner/pyproject.toml` already permits it (`aiohttp>=3.14.1`), so only `runner/uv.lock` moves — 3.14.1 → 3.14.3, resolved as a single-package upgrade with nothing else shifting.

The runner declares `aiohttp` directly but does not import it anywhere in its own source; it is reached through provider SDKs acting as HTTP clients, which is the direction the response-parser advisory applies to.

## Verification

- `uv lock --upgrade-package aiohttp` → 3.14.3, single package changed
- `uv sync` + `import aiohttp` confirms 3.14.3 installed
- Full runner suite: **1247 passed, 2 skipped** (the 2 skips are pre-existing — optional `google-stt` / `google-tts` extras not installed)

## Not in this PR

`benchmarks` also has near-deadline npm alerts in a different manifest — #48 `sharp` (→0.35.0, ~17 days) and #58 `postcss` (→8.5.18, ~19 days), plus #38 `dompurify`. Those are a separate lockfile and a separate change.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


ben-coval (approved):

callumreid (approved):
Approving. Lockfile-only aiohttp 3.14.1 → 3.14.3, matching backend#6437 so the two repos stay on the same version of a security-relevant dependency rather than drifting.

Single file, hashes and wheel URLs regenerated by the resolver rather than hand-edited, green.

PR #451, opened 2026-08-05
chole(deploy): publish verified Benchmarks deployment state

borgesius:
## What this changes

Benchmarks Web and runner releases emit the same structured deployment evidence used by the shared promotion system.

- Web records its immutable source SHA and Vercel deployment URL.
- Runner records its immutable source SHA and verified infrastructure rollout.
- Only successful manual releases seed the verified service/environment baseline.
- Automatic releases retain evidence but do not seed or move state.
- No automatic-promotion guard is activated in this PR.

The Web workflow also runs the shared Vercel action from the actual web package directory, including code generation before build.

## Dependencies

The shared promotion foundation in [Switchboard #406](https://github.com/coval-ai/ci-cd-switchboard/pull/406) and canonical adapter context in [Switchboard #414](https://github.com/coval-ai/ci-cd-switchboard/pull/414) are merged. Benchmarks intentionally keeps its private shared-action checkout pinned to immutable reviewed Switchboard merge commit 84d68f1f389588d9403971bced2160ede7b0c49c.

## Validation

Current head 9cecd5c is based on current Benchmarks main.

- deployment-state and Web-directory workflow tests: 5 passed
- complete Web unit suite: 116 passed
- Web typecheck and lint: passed
- actionlint: passed
- earlier hosted pointed validation: [run 31051510591](https://github.com/coval-ai/benchmarks/actions/runs/31051510591)
- current exact-head hosted validation is required before merge
- no production release, deployment, promotion, or baseline publication was triggered


coval-cale (approved):

.github/workflows/web-release.yml
```diff
@@ -47,20 +47,28 @@ jobs:
       deploy_ref: ${{ steps.release.outputs.deploy_ref }}
```
callumreid, line 47: **P0 - this workflow no longer exists on `main`; the Web half of the PR edits a deleted file.**

`gh api compare 102302e0da6e60cf6db1b019626ed0c78841fd27...main` reports `.github/workflows/web-release.yml` with status `removed`, along with `.github/workflows/web-ci.yml` and every `web/**` path. The deletion is commit `19ac5e18` (2026-08-07), "Delete the frontend, which now lives in benchmarks-web (#469)". `main`'s root tree has no `web/`, and `main`'s workflow directory has no `web-release.yml`. This branch's merge base is `102302e0` and it is `behind_by: 33`.

Failure mode: this is a modify/delete against `main`. Either the merge conflicts, or — if it is resolved by taking the branch side — it resurrects a production release workflow for a frontend that has been moved out of this repo, alongside a new `web/lib/*.test.ts` under a directory `main` deleted (including its `vitest.config.ts`), so the test has no runner and would read a workflow file that only exists because the merge put it back.

Why existing checks miss it: GitHub computes the diff against the merge base, so the deleted-on-main files simply do not appear as conflicts in the diff view; `mergeable` is still `UNKNOWN` on this draft. `Benchmarks Web CI Adapter` reports `skipping`, and the PR's local validation ("5 passed", "116 passed") ran against the stale pre-deletion tree, which is also why the body's "current head 9cecd5c is based on current Benchmarks main" reads as true locally but is not.

Smallest correction: rebase onto current `origin/main` and drop the Web changes from this PR entirely — the `web-release.yml` edits (Switchboard SHA bump to `84d68f1f`, `vercel-directory: web`, the record/upload/publish steps) belong in `coval-ai/benchmarks-web`, whose `web-release.yml` is still on `9221d6f4` with `install-directory: .` / `vercel-directory: .`. Note that repo's package is at its own root, so `vercel-directory` there should stay `.`, not become `web`. What remains here is the runner half, which is clean: `runner-release.yml` is untouched on `main` since the merge base and merges without conflict.

web/lib/deploymentStateWorkflows.test.ts
```diff
@@ -0,0 +1,69 @@
+import fs from "node:fs";
```
callumreid, line 1: **P2 - the test asserts YAML substrings, so it cannot fail for the reasons that would actually break a release.**

Every assertion is `expect(<raw yaml string>).toContain(<literal from that same yaml>)`. It restates the file rather than exercising it: it would still pass if `promotion_origin` were wired to the wrong job output, if `record_deployment`'s `if:` gate inverted, or if the emitted JSON stopped validating against `switchboard-deploy-result-v1`. Two assertions are additionally brittle rather than semantic — `expect(webRelease).not.toContain("cd web\n            pnpm codegen")` encodes an exact 12-space indent and silently stops testing anything if the block is reindented, and `promotionFoundationSha` duplicates a constant whose real source of truth is the Switchboard merge commit.

Why existing checks miss it: it is green today precisely because it mirrors the file, so it reads as coverage while providing none.

Smallest correction: when this lands in `coval-ai/benchmarks-web`, parse the workflow with a YAML loader and assert on the resolved structure (job/step `if` expressions, the publish step's `with:` inputs), or — higher value for the effort — run the `jq -n` payload builder with fixture env values and validate the output against `switchboardDeployResultSchema`, which is the invariant that actually gates publication.

.github/workflows/runner-release.yml
```diff
@@ -39,15 +39,54 @@ jobs:
...
+              echo "::error::promotion_origin must be automatic or manual."
+              exit 1
+              ;;
+          esac
+          selected_deploy_action_ids="$(
+            jq -c '
+              if type == "array" and length > 0
+              then .
+              else ["benchmarks.runner.production.release"]
+              end
```
callumreid, line 77: **P3 - runner silently substitutes a default for an explicitly empty selection, where web errors.**

`if type == "array" and length > 0 then . else ["benchmarks.runner.production.release"] end` collapses two different dispatch payloads into one: `selected_deploy_action_ids` absent (`null`) and `selected_deploy_action_ids: []`. The web workflow at `web-release.yml` deliberately distinguishes them — its `if type == "array" then .` keeps `[]` as `[]`, and the guard immediately below then fails with "The Benchmarks Web production release action was not selected."

Failure mode: a Switchboard dispatch that explicitly deselects the runner release action (empty array) is treated as if the action had been selected. The runner deploys and, on `promotion_origin: manual`, publishes a verified baseline whose `selected_deploy_action_ids` claims an action the caller did not choose. `deployResult.ts` cannot catch this — it only checks that selected IDs and reported results agree with each other, which they do.

Why existing checks miss it: the guard on the next line is unreachable for `[]`, because the default has already been substituted by then.

Smallest correction: drop `and length > 0` so the shape matches web — absent still defaults, explicit `[]` falls through to the existing guard and fails loudly.

callumreid (changes requested):
## What this does

Teaches both Benchmarks release workflows to emit a `switchboard-deploy-result-v1` artifact and, on a successful **manual** production release, publish it as the service's verified promotion baseline via the shared Switchboard action. It also accepts and validates `mode` / `promotion_origin` / `selected_deploy_action_ids` from the dispatch payload, bumps the pinned Switchboard checkout, and repoints the Web Vercel build at the `web/` package.

## What I verified against the trees

The Switchboard-side contract checks out clean. At the pinned commit `84d68f1f`:

- `config/switchboard/04-services.yaml` declares `benchmarks.web` and `benchmarks.runner` with `adapter.supportedModes: [safe, plan, deploy_only]` — exactly the set both workflows accept — plus `resultContract: switchboard-deploy-result-v1` and `resultEnforcement: shadow`.
- `config/switchboard/08-deploy-actions.yaml` contains `benchmarks.web.production.release` and `benchmarks.runner.production.release` verbatim, so the guard strings are real IDs, not invented ones.
- `switchboard-publish-deployment-state/action.yml` takes exactly the five inputs passed here (`environment`, `repository`, `result-path`, `service`, `token`) and pushes with that token, which is why the runner's `record_deployment` job correctly declares `contents: write`.
- Both emitted JSON payloads validate against `src/switchboard/deployResult.ts`: `built_artifact.kind` values `other` / `vercel_deployment` are in the enum, `built_artifact.identity == live_artifact.identity` in both, every selected action has a matching result, and `evidence_urls` only picks up `infra_url` / `deployment_url` on the `deployment_performed` branch (they're `z.string().url()`, so including them unconditionally would have failed).
- `vercel-release/action.yml` runs `pre-build-command` with `working-directory: ${{ inputs.vercel-directory }}`, so moving `vercel-directory` to `web` and dropping the `cd web` is the correct pairing — `.vercel/.env.*.local` now resolves next to where `vercel pull` ran. Its `deployed` output is `deployment_url != ''`, i.e. literal `true`/`false`, which is what `--argjson deployment_performed` needs.

The runner half is sound and merges cleanly (`runner-release.yml` is untouched on `main` since the merge base). CI on this head is green.

## Why this isn't an approve

The Web half of this PR edits files that no longer exist in this repository. `git compare 102302e0...main` shows `.github/workflows/web-release.yml`, `.github/workflows/web-ci.yml`, and all of `web/**` with status `removed` — deleted by "Delete the frontend, which now lives in benchmarks-web (#469)" on 2026-08-07. This branch is 33 commits behind and still carries the pre-deletion tree, so the PR body's "current head 9cecd5c is based on current Benchmarks main" does not hold, and the local "5 passed / 116 passed" runs exercised a `web/` package that `main` doesn't have. Detail and the smallest correction are inline.

The runner-side work looks ready to land on its own once it's rebased.

PR #459, opened 2026-08-06
[BENCH-632] Runner: capture native usage quantities per result (1/6)

cooperr1:
First of the six stacked BENCH-631 cost/pricing PRs — merge in ticket order; each PR diff is only its own commit.

Widens `TranscriptionResult`/`TTSResult` with raw usage fields and parses them where providers already report usage (Mistral + OpenAI Realtime tokens; Inworld/xAI/Deepgram/Cartesia/Modulate/Gradium billable duration — Deepgram sums is_final frame durations, Cartesia/Gradium are cumulative so last wins). The orchestrator stamps every result row with known quantities (STT `audio_seconds_in` from the trimmed item, TTS `characters_in` + measured `audio_seconds_out`) and meters the whisper-1 judge per run. Migration `0015` adds the nullable columns (plus `judge_audio_seconds` beyond the ticket DDL: whisper-1 reports duration-type usage, its actual billing unit). No behavior change to metrics; all fields nullable; providers that report nothing still work.

Tests: provider fixtures now assert parsed usage; real-Postgres migration up/down + writer roundtrip; judge accumulation covered for both usage shapes.

<h3>Greptile Summary</h3>

Adds nullable native usage quantities to benchmark result and run records.
- Parses token or billed-duration usage reported by supported STT providers.
- Stamps STT and TTS metric rows with request-level input/output quantities.
- Accumulates Whisper judge usage and persists it when finalizing a run.
- Adds the corresponding migration, database writer support, fixtures, and tests.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge; no concrete changed-code failure remains supported by the available repository evidence.

The migration, persistence models, writer SQL, provider parsing, and orchestrator propagation are coordinated, while investigated edge cases either match the documented design or lack a reachable failing condition.

<sub>Reviews (1): Last reviewed commit: ["\[BENCH-632\] Runner: capture native usage..."](https://github.com/coval-ai/benchmarks/commit/2ce6c146dea2d3f0d8c40ac610f4b5d9f264d1f8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50942993)</sub>

**Context used:**

- Knowledge Base — [Runner Orchestration](https://app.greptile.com/coval/-/custom-context/knowledge-base/coval-ai/benchmarks/-/docs/runner-orchestration.md)
- Knowledge Base — [DB Layer](https://app.greptile.com/coval/-/custom-context/knowledge-base/coval-ai/benchmarks/-/docs/db-layer.md)

PR #460, opened 2026-08-06
[BENCH-633] Pricing store: append-only model_pricing table + seed ratesheet (2/6)

cooperr1:
Stacked on #459.

Append-only `model_pricing`: one effective rate per (provider, model, benchmark, billing_unit) enforced by a partial unique index — benchmark is in the key because gradium serves STT and TTS under one model id. `PricingStore` does the supersede dance atomically, is idempotent on unchanged rates, and serves a run-scoped cache. The seed prices 62/68 active models with per-page provenance (source_url, as_of, verbatim evidence, AA-style plan assumptions) plus the whisper-1 judge; the 6 unpriced models are an explicit gap list that `check_pricing_coverage.py` reports. Every rate was verified against the provider's public pricing page on 2026-08-06; deepgram/elevenlabs/openai/assemblyai independently re-verified before opening this PR.

<h3>Greptile Summary</h3>

The PR adds an append-only model-pricing table, typed pricing store, local seed ratesheet, coverage checker, and database-backed tests.

- Adds historical rate supersession and current-rate caching.
- Seeds published pricing with provenance and explicit unpriced-model gaps.
- Adds migration and tests for uniqueness, history, token-billed rates, caching, and seed idempotency.

The benchmark-unscoped pricing reads need to be fixed before merging because shared provider/model identities return rates from multiple benchmarks.

The seeded data already contains `gradium/default` under both STT and TTS, while both database and cached reads omit benchmark and therefore return incompatible rates together.

**Files Needing Attention:** runner/src/coval_bench/db/pricing.py

<sub>Reviews (1): Last reviewed commit: ["\[BENCH-633\] Pricing store: append-only m..."](https://github.com/coval-ai/benchmarks/commit/d06594aa189ec869960a543283ffab706eac9f90) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50943004)</sub>

**Context used:**

### DIFF
--- a/runner/scripts/check_pricing_coverage.py
+++ b/runner/scripts/check_pricing_coverage.py
@@ -28,10 +28,7 @@ async def _unpriced() -> list[str]:
             f"{m.benchmark}:{m.provider}/{m.model}"
             for m in MODEL_REGISTRY
             if m.status in (ModelStatus.ACTIVE, ModelStatus.EARLY_ACCESS)
-            and not any(
-                r.benchmark is m.benchmark
-                for r in store.effective_rates_cached(m.provider, m.model)
-            )
+            and not store.effective_rates_cached(m.provider, m.model, m.benchmark)
         ]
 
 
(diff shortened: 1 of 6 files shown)


cooperr1:
Greptile's P1 is half-right: every existing caller (cost calculator, normalizer, coverage check) already filtered by benchmark on the returned rows, so no incompatible rate could actually be applied. But the footgun for future callers was real — 70983a2 pushes the benchmark filter into the store itself (`get_effective_rates`/`effective_rates_cached` take an optional `benchmark`, SQL + cache), converts the coverage check to it, and adds a gradium/default test proving STT and TTS rates never co-mingle.

PR #461, opened 2026-08-06
[BENCH-634] Cost computation: COST_USD per result + per-run rollups (3/6)

cooperr1:
Stacked on #460.

Cost as a metric: each succeeded item with a resolvable rate gets a sibling `COST_USD` row (token rates × counts → duration rates × billable/measured seconds → per-1M-chars; no rate or quantity → no row, logged once per model per run — never a zero). `finish_run` sums the run's persisted cost rows in SQL (correct even after SIGTERM) plus the whisper judge's cost into `runs.total_cost_usd`/`judge_cost_usd` (migration `0017`), and a fail-silent `benchmark_run_cost` PostHog event carries totals + top-5 model costs. `COST_USD` is internal-only: `INTERNAL_METRICS` filters it from `/v1/results` and both aggregates endpoints (API test proves it) while it deliberately stays in `results_by_bucket` for internal spend series. Canonical queries in `docs/cost-tracking.md`.

<h3>Greptile Summary</h3>

This PR adds per-item `COST_USD` measurements, persisted per-run cost rollups, internal-only API filtering, and PostHog spend telemetry.

- Computes costs from effective token, duration, or character rates.
- Stores item and judge costs and rolls them into `runs.total_cost_usd`.
- Keeps cost metrics out of public result and aggregate responses while retaining internal bucket data.
- Adds cost calculation, persistence, API, registry, and orchestration tests.

<h3>Confidence Score: 3/5</h3>

The PR should not merge until PostHog cost telemetry uses persisted totals and covers SIGTERM-finalized runs.

The database rollup is authoritative, but normal telemetry can include rows whose persistence failed, while SIGTERM runs can persist spend without emitting any cost event.

**Files Needing Attention:** runner/src/coval_bench/runner/orchestrator.py

<sub>Reviews (1): Last reviewed commit: ["\[BENCH-634\] Cost computation: COST\_USD p..."](https://github.com/coval-ai/benchmarks/commit/37bb0a8dc8ae71f6e2fe79f754dd841bf56d132d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50943165)</sub>

> Greptile also left **2 inline comments** on this PR.

<details><summary><h4>Context used (3)</h4></summary>

- Knowledge Base — [Runner Orchestration](https://app.greptile.com/coval/-/custom-context/knowledge-base/coval-ai/benchmarks/-/docs/runner-orchestration.md)
[...]

PR #462, opened 2026-08-06
[BENCH-635] Price collector: cron job with LLM extraction, diff + review gate (4/6)

cooperr1:
Stacked on #461.

`coval-bench update-prices`: scrapes every provider page in the new `pricing_sources` registry (explicit None for the pageless), snapshots pages (gzipped, hash-deduped) so bot rows carry reviewable evidence, extracts rates via OpenAI structured outputs under an extract-only-never-infer prompt with mandatory verbatim quotes, matches to registry ids (exact + alias map; ambiguous ids never guessed), cross-checks LiteLLM as a second witness, and gates: identical → noop; ≤20% delta + high confidence + witness agrees-or-unavailable → auto-apply as `updated_by='bot'`; anything else (including brand-new rates) → exactly one review item via Linear or Slack webhook. Staleness sweep flags rates >45 days; one provider failing never aborts the rest; `--dry-run` prints the diff table and writes nothing. Weekly cadence mirrored as `PRICING_UPDATE_PERIOD_SECONDS` (cron lives in benchmark-infra).

<h3>Greptile Summary</h3>

The PR adds a weekly pricing collector that snapshots provider pages, uses structured LLM extraction, gates price changes against current rates and LiteLLM, writes approved rates, and sends uncertain or stale cases for review.

- Adds the `update-prices` CLI and collector workflow.
- Adds pricing-source and model-alias registries.
- Adds persisted, gzipped pricing-page snapshots.
- Adds configuration, documentation, fixtures, and integration tests for collection and review behavior.

<h3>Confidence Score: 0/5</h3>

The PR is not safe to merge until dry runs become read-only, successful unchanged checks refresh staleness, unverified LLM output cannot auto-apply pricing, and unmatched rates reach review.

The current workflow can write during dry runs, repeatedly flag freshly verified rates as stale, persist pricing based on self-attested extraction when the independent witness is absent, and omit newly discovered models from the review queue.

**Files Needing Attention:** runner/src/coval_bench/pricing/collector.py, runner/src/coval_bench/__main__.py, runner/tests/pricing/test_collector.py

<details open><summary><h3>Security Review</h3></summary>

The auto-apply path does not verify that the extractor's quote and rate are present in the fetched page, while an unavailable LiteLLM witness is accepted. Prompt-influenced page content can therefore satisfy the gate and alter effective pricing used for cost calculations. **How this was verified:** The parsed extraction flows directly to `upsert_rate`, and the only witness condition that blocks it is an explicit LiteLLM disagreement.
</details>

<sub>Reviews (1): Last reviewed commit: ["\[BENCH-635\] Price collector: cron job wi..."](https://github.com/coval-ai/benchmarks/commit/d0157d4d4e9395cbedff0f49c207946dc06d444c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50943174)</sub>

> Greptile also left **4 inline comments** on this PR.

**Context used:**

### DIFF
--- a/runner/README.md
+++ b/runner/README.md
@@ -4,6 +4,15 @@ Runner + API for Coval voice-AI benchmarks. Implements:
 
 - A Cloud Run Job that runs STT/TTS providers against a pinned dataset every 30 min and writes results to Cloud SQL.
 - A FastAPI service that serves the public results API at `https://benchmarks.coval.ai`.
+- A weekly price-collector Cloud Run Job (`coval-bench update-prices`, same
+  image, different command; cron in `benchmark-infra`, cadence mirrored by
+  `PRICING_UPDATE_PERIOD_SECONDS`). It scrapes provider pricing pages,
+  LLM-extracts rates with verbatim quotes, snapshots the pages for evidence,
+  auto-applies small agreed changes to `benchmarks_v2.model_pricing` as
+  `updated_by='bot'` rows, and opens a review item (Linear via
+  `LINEAR_API_KEY`, or `PRICING_REVIEW_SLACK_WEBHOOK`) for large or uncertain
+  ones and for rates stale beyond 45 days. `--dry-run` prints the diff table
+  without writing.
 
 ## What we benchmark
 
--- a/runner/src/coval_bench/__main__.py
+++ b/runner/src/coval_bench/__main__.py
@@ -100,6 +100,53 @@ migrate.add_command(import_legacy_cli, name="import-legacy")
 cli.add_command(fetch_s2s, name="fetch-s2s")
 
 
+@cli.command(name="update-prices")
+@click.option(
+    "--dry-run",
+    is_flag=True,
+    default=False,
+    help="Print the diff table only — write no snapshots, rates, or review items.",
+)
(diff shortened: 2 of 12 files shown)


cooperr1:
All four findings were legitimate — fixed in 93ceecc:

- **Unverified extraction auto-applies (security)**: auto-apply is now gated on the extractor's verbatim quote being found in the fetched page (tag-stripped, whitespace-collapsed); the prompt requires character-for-character quotes with derivations moved to a separate `note` field, and a new test proves a fabricated quote goes to review even inside the delta gate. LiteLLM-unavailable alone is no longer sufficient.
- **Dry-run persists snapshots**: `--dry-run` now computes the hash without inserting; test asserts zero snapshot rows.
- **No-op leaves verification stale**: a re-verified unchanged rate gets today's `as_of` stamped (the rate itself is untouched — verification metadata only, same category as `superseded_at`), so the staleness alarm can't fire right after a successful check.
- **Unmatched rates bypass review**: unmatched extractions now aggregate into the alert review item (one line per rate, never auto-written) instead of dying in logs.

PR #469, opened 2026-08-07
Delete the frontend, which now lives in benchmarks-web

coval-cale:
Drops web/ and its two workflows, the switchboard web adapter and its full-CI fallback, the advisory methodology marker check (markers are updated by hand now), and the web half of arena-local.sh.


cooperr1 (approved):

runner/scripts/methodology_marker_check.py
```diff
@@ -1,96 +0,0 @@
...
-
-from __future__ import annotations
-
-import argparse
-import shutil
-import subprocess
-import sys
-from collections.abc import Iterable, Sequence
-
-METHODOLOGY_MARKER_FILE = "web/lib/config/methodologyChanges.ts"
```
seribaymadina, line 12: do you feel like we dont need methodology marker anymore? dont seem to locate it in a frontend repo

coval-cale: I think we def still need it. But we also need to figure out where that goes. Right now we still have all the methodology change info in `benchmarks-web` but I'm thinking of moving it to the backend so its at the source of where the changes are.

web/scripts/coval-dev-prod
```diff
@@ -1,94 +0,0 @@
...
-fi
-
-# --- detect the worktree you're currently in ---
-ROOT="$(git -C "$PWD" rev-parse --show-toplevel 2>/dev/null || true)"
-if [ -z "$ROOT" ]; then
-  echo "✗ Not inside a git worktree. cd into a coval-benchmarks worktree first." >&2
-  exit 1
-fi
-WEB_DIR="$ROOT/web"
-if [ ! -d "$WEB_DIR" ]; then
```
seribaymadina, line 32: this seems to assume it's still in the old monorepo, hence broken

scripts/arena-local.sh
```diff
@@ -1,10 +1,11 @@
 #!/usr/bin/env bash
-# Local Voice Arena stack: Postgres + API (docker compose) + Next.js web (pnpm).
```
seribaymadina, line 2: this script can't be run with one command anymore; feels like it's easier to just delete it fully and rewrite it in accordance with the split ( if necessary)

web/scripts/smoke-playground.sh
```diff
@@ -1,126 +0,0 @@
...
-# Exercises the security + validation ladder against a running dev server using
-# FAKE provider keys — no real provider calls succeed, so no spend is incurred.
-# The auth/validation checks (403/401/413/400) are deterministic regardless of
-# keys; the upstream checks only assert the request reached the provider stage.
-#
-# Usage:
-#   1. In web/.env.local set PLAYGROUND_SESSION_SECRET to any 32+ byte base64
-#      value and the PLAYGROUND_*_API_KEY vars to fake values (e.g. "fake-key").
-#   2. Run the dev server:  pnpm dev
-#   3. In another shell:    bash web/scripts/smoke-playground.sh
```
seribaymadina, line 14: nit: have to get rid of web part :  
.env.local
bash scripts/smoke-playground.sh

web/scripts/coval-dev-prod
coval-cale: Nice catch! Pushing a fix!

scripts/arena-local.sh
coval-cale: Nice I'll add a ticket to do that.

.github/workflows/web-ci.yml
```diff
@@ -1,104 +0,0 @@
...
-          python-version: "3.12"
-
-      - name: Install dependencies
-        run: pnpm install --frozen-lockfile
-
-      - name: Install runner dependencies for OpenAPI schema
-        run: uv sync --frozen --dev
-        working-directory: runner
-
-      - name: Serve local OpenAPI schema
```
seribaymadina, line 70: still trying to wrap my head around a way to do it but this seems valid: the old CI generated the frontend schema from the backend code in the same checkout. New frontend CI generates against the deployed production API. Therefore, a benchmarks-coval PR can change or remove an API field and still pass without testing benchmarks-web.

web/lib/arena/domains.ts
```diff
@@ -1,15 +0,0 @@
-// Values must match ArenaDomain in runner/src/coval_bench/api/schemas.py — they are
```
seribaymadina, line 1: can drift? need some contract test or

.github/workflows/web-ci.yml
coval-cale: Ya super sick idea. I think this is the whole point of Dana's switchboard. I'll look into it.

web/lib/arena/domains.ts
coval-cale: Will have to be added to the codegen contract.

PR #483, opened 2026-08-10
Add Inworld TTS 2 Flash under early access

seribaymadina:
Adds `inworld-tts-2-flash`, Inworld's latency-tuned sibling to TTS 2, under early access 

Note:
- Inworld's docs also mark `inworld-tts-1.5-max` and `-mini` deprecated, but the automatic rerouting runs into the 1.5 family rather than out of it, so those entries stay active and keep reporting themselves. 
- Web side is in diff repo

<h3>Greptile Summary</h3>

Adds Inworld TTS 2 Flash to the existing WebSocket provider and registers it as an early-access TTS model.
- Extends the Inworld provider model allowlist.
- Registers Brooke and Jason as the model's benchmark voices.
- Adds constructor-level coverage for the new model identifier.

The scheduling configuration should be corrected before merging because the new model will run on every normal shared benchmark rather than only nightly.

The early-access status makes the model schedulable, while its default shared source routes it into normal runs; repository nightly examples explicitly use the dedicated source.

**Files Needing Attention:** runner/src/coval_bench/registries/models.py

<sub>Reviews (1): Last reviewed commit: ["Add Inworld TTS 2 Flash under early acce..."](https://github.com/coval-ai/benchmarks/commit/b549a42efc1b2f0cb356436e70e4633803fabb0b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51976138)</sub>

**Context used:**

### DIFF
--- a/runner/src/coval_bench/providers/tts/inworld.py
+++ b/runner/src/coval_bench/providers/tts/inworld.py
@@ -44,7 +44,14 @@ def _pcm_from_chunk(chunk: bytes) -> bytes:
 class InworldTTSProvider(TTSProvider):
     """Inworld AI TTS provider using WebSocket bidirectional streaming."""
 
-    _VALID_MODELS = frozenset({"inworld-tts-2", "inworld-tts-1.5-max", "inworld-tts-1.5-mini"})
+    _VALID_MODELS = frozenset(
+        {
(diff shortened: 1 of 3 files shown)


coval-cale (approved):
LGTM!

PR #485, opened 2026-08-10
Pair arena battles within one gender

seribaymadina:
Both sides of a battle now speak the same gender, so a vote compares delivery rather than register. Whichever gender has fewer battles goes next, keeping the two vote counts even.

Battles record the voices that sang and the gender they shared, so a rating stays traceable to a speaker. Voice ids stay out of every API response including reveal — an id like `aura-2-orion-en` names its provider.

Ratings continue rather than restarting. Votes from the cross-gender era still count, so the board keeps its history and the older share dilutes as same-gender votes land. Legacy battles are no longer handed out for voting though — their votes count, but serving one would keep adding the comparison this change stopped making.

The methodology version moves to `davidson-bt-002` to mark the date pairing changed, not a reset. Palabra has no gendered voices, so it sits out until handled separately.

### DIFF
--- a/runner/src/coval_bench/api/routers/arena.py
+++ b/runner/src/coval_bench/api/routers/arena.py
@@ -63,7 +63,8 @@ from coval_bench.arena.monitoring import (
 from coval_bench.arena.pairing import (
     PAIRING_DOMAIN,
     PAIRING_METRIC,
-    active_tts_models,
+    gender_for_next_battle,
+    roster_for,
     select_pair,
 )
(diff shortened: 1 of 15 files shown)


seribaymadina:
Deliberate, since current results didn't have many votes per provider slight dilution of votes won't be an issue + "deficit-based" selection self-corrects: a concurrent double-pick  of the same gendered pairing overshoots by one, and the next battle sees the imbalance and picks the other gender, so drift never latches the way a parity bit would.

coval-cale (approved):
Lets GOOOOO!

PR #495, opened 2026-08-12
[BENCH-667] Normalize benchmark metric storage

kdmelon:
## Summary

- add an additive normalized benchmark schema for observations, preprocessing artifacts, metric evaluations, metric values, metric artifacts, and generic metric rollups
- support multiple timestamp producers and models per observation through explicit producer, provider, model, and version provenance
- support named evaluation variants such as `google`, `deepgram`, and `ensemble` without conflating their results
- freeze each evaluation's exact, ordered, role-aware preprocessing inputs when it is queued
- preserve existing `runs`, `results`, and `results_by_bucket`; this PR does not cut existing readers or writers over

## Design notes

- preprocessing remains separate from metric evaluation: timestamp artifacts are immutable private GCS pointers, while metric evaluations reference the exact artifacts they consume
- artifact retry identity includes the observation, schema version, pipeline version, and complete producer identity
- evaluation identity and input lineage are immutable after creation; exact retries return the existing row and conflicting retries fail
- database guards require linked artifacts to belong to the same observation and allow cleanup only through legitimate parent cascades
- metric values remain long-form rows with one unit and an explicit primary value
- `evaluation_variant` is included in evaluation uniqueness and rollup keys so provider-specific and ensemble results cannot mix
- successful metric output is validated atomically

## Scope boundaries

- no GPU timestamp worker, Cloud Run job, private manifest, or model inference is added here; those remain in BENCH-668
- no pronunciation scoring is included; that remains private CPU work under BENCH-669
- no dataset contents or transcripts are stored in the public repository
- no backend cutover is included

## Verification

- Ruff lint: pass
- Ruff format check: pass
- strict mypy for normalized DB code and focused tests: pass
- focused real-PostgreSQL migration, writer, provenance, lifecycle, cascade, retry, and rollup tests: 17 passed
- regression coverage verifies evaluation identity cannot change during lifecycle transitions, unrelated nested triggers cannot bypass immutable lineage, and legitimate parent cascades remove linked inputs

Linear: https://linear.app/coval/issue/BENCH-667/normalize-benchmark-storage-for-evolving-metrics


runner/src/coval_bench/db/migrations/versions/20260813_0018_normalized_benchmark_storage.py
```diff
@@ -0,0 +1,177 @@
...
+            value_key TEXT NOT NULL CHECK (value_key <> ''), unit TEXT NOT NULL CHECK (unit <> ''), bucket_at TIMESTAMPTZ NOT NULL,
+            min_value DOUBLE PRECISION NOT NULL CHECK (min_value NOT IN ('NaN'::float8, 'Infinity'::float8, '-Infinity'::float8)), p25 DOUBLE PRECISION NOT NULL CHECK (p25 NOT IN ('NaN'::float8, 'Infinity'::float8, '-Infinity'::float8)), p50 DOUBLE PRECISION NOT NULL CHECK (p50 NOT IN ('NaN'::float8, 'Infinity'::float8, '-Infinity'::float8)), p75 DOUBLE PRECISION NOT NULL CHECK (p75 NOT IN ('NaN'::float8, 'Infinity'::float8, '-Infinity'::float8)), max_value DOUBLE PRECISION NOT NULL CHECK (max_value NOT IN ('NaN'::float8, 'Infinity'::float8, '-Infinity'::float8)), value_sum DOUBLE PRECISION NOT NULL CHECK (value_sum NOT IN ('NaN'::float8, 'Infinity'::float8, '-Infinity'::float8)), sample_count INTEGER NOT NULL CHECK (sample_count > 0),
+            CHECK (min_value <= p25 AND p25 <= p50 AND p50 <= p75 AND p75 <= max_value),
+            PRIMARY KEY (provider, model, benchmark, dataset_id, metric_type, metric_version, value_key, bucket_at)
+        );
```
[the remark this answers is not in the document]
kdmelon: Fixed in 8c4be1a: added the bucket_at index and downgrade coverage.

runner/src/coval_bench/db/models.py
```diff
@@ -30,9 +31,267 @@
...
+
+def _private_gs_uri(value: str) -> str:
+    if not value.startswith("gs://") or "?" in value or "#" in value:
+        raise ValueError("must be a private gs:// URI without query or fragment")
+    return value
+
+
+def _optional_private_gs_uri(value: str | None) -> str | None:
+    """Accept absent source audio while validating any supplied object URI."""
+    return None if value is None else _private_gs_uri(value)
```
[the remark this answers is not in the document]
kdmelon: Fixed in 8c4be1a: Pydantic and PostgreSQL now both require a bucket plus object path, and preprocessing duration must be finite.

runner/src/coval_bench/db/writer.py
```diff
@@ -146,6 +156,461 @@ async def record_results(self, results: Sequence[Result]) -> None:
...
+                            observation.provider,
+                            observation.model,
+                            observation.voice,
+                            observation.benchmark,
+                        ),
+                    )
+                    row = await cur.fetchone()
+            await conn.commit()
+        if row is None:  # pragma: no cover
+            raise RuntimeError("INSERT INTO benchmark_observations returned no row")
```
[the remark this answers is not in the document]
kdmelon: Keeping the existing flow. Under PostgreSQL READ COMMITTED, ON CONFLICT waits for the competing insertion and the subsequent SELECT gets a fresh statement snapshot. A no-op UPDATE would introduce unnecessary writes and trigger exposure.

runner/src/coval_bench/db/writer.py
```diff
@@ -146,6 +156,461 @@ async def record_results(self, results: Sequence[Result]) -> None:
...
+                    "SELECT pg_advisory_xact_lock(hashtextextended('metric_values_by_bucket',"
+                    " extract(epoch FROM %(bucket)s::timestamptz)::bigint))",
+                    params,
+                )
+                await cur.execute(
+                    "DELETE FROM benchmarks_v2.metric_values_by_bucket "
+                    "WHERE bucket_at = %(bucket)s",
+                    params,
+                )
+                await cur.execute(
```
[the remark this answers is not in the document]
kdmelon: Fixed in 8c4be1a: removed period_seconds from the normalized rollup and updated callers.

runner/tests/unit/test_normalized_db_writer.py
```diff
@@ -0,0 +1,620 @@
...
+            (evaluation_id,),
+        )
+        cur.execute(
+            """UPDATE benchmarks_v2.metric_evaluations
+               SET status = 'succeeded', finished_at = now() WHERE id = %s""",
+            (evaluation_id,),
+        )
+        with pytest.raises(psycopg.errors.RaiseException, match="all components"):
+            cur.execute("COMMIT")
+        cur.execute("ROLLBACK")
```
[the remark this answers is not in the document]
kdmelon: Fixed in 8c4be1a: added direct non-queued insertion coverage, plus broader lifecycle and contract tests.

coval-cale:
Taking a look rn. Might take a bit to fully go through. Just want to make sure this doesn't kill benchmarks before the AMA tomorrow. Thanks!

kobicovaldev (commented):
Test review

runner/src/coval_bench/db/models.py
```diff
@@ -32,9 +33,279 @@
...
+    audio_duration_ms: int | None = Field(default=None, gt=0)
+    captured_at: datetime | None = None
+    status: ObservationStatus
+    error: str | None = None
+
+    _validate_audio_uri = field_validator("audio_uri")(_optional_private_gs_uri)
+
+    @model_validator(mode="after")
+    def _outcome_matches_error(self) -> Observation:
+        import math
```
kobicovaldev: [Suggestion] Would it be cleaner to move `import math` to the module-level imports alongside `datetime` and `UUID`? It is lazily imported in four places (`_outcome_matches_error`, `_artifact_identity`, `_finite`, and `_valid_percentiles`), and the same pattern appears in `registries/metrics.py`. A single top-level import would be consistent and expose any missing dependency upfront.

runner/src/coval_bench/db/writer.py
```diff
@@ -146,6 +156,461 @@ async def record_results(self, results: Sequence[Result]) -> None:
...
+                       WHERE id = %s AND status = %s
+                       RETURNING id, observation_id, metric_type, metric_version, executor,
+                                 external_request_id, status, started_at, finished_at, error,
+                                 attempt_count, created_at, updated_at""",
+                    (ProcessingStatus.RUNNING, started_at, evaluation_id, ProcessingStatus.QUEUED),
+                )
+                row = await cur.fetchone()
+            await conn.commit()
+        if row is None:
+            raise ValueError("only queued metric evaluations may be started")
```
kobicovaldev: [Required] This error message does not include `evaluation_id`, so it is hard to diagnose which evaluation hit this path. Would it be clearer to include it, like `f"metric evaluation {evaluation_id} is not queued"`?

runner/src/coval_bench/db/writer.py
```diff
@@ -146,6 +156,461 @@ async def record_results(self, results: Sequence[Result]) -> None:
...
+                        error,
+                        evaluation_id,
+                        ProcessingStatus.QUEUED,
+                        ProcessingStatus.RUNNING,
+                    ),
+                )
+                row = await cur.fetchone()
+            await conn.commit()
+        if row is None:
+            raise ValueError("only queued or running metric evaluations may fail")
```
kobicovaldev: [Required] This error message does not include `evaluation_id`. Would it be clearer to include it, like `f"metric evaluation {evaluation_id} is not queued or running"`?

runner/src/coval_bench/db/writer.py
```diff
@@ -146,6 +156,461 @@ async def record_results(self, results: Sequence[Result]) -> None:
...
+        async with self._pool.connection() as conn:
+            async with conn.cursor(row_factory=psycopg.rows.dict_row) as cur:
+                await cur.execute(
+                    """SELECT metric_type, metric_version, status, finished_at
+                       FROM benchmarks_v2.metric_evaluations WHERE id = %s FOR UPDATE""",
+                    (evaluation_id,),
+                )
+                evaluation = await cur.fetchone()
+                if evaluation is None:
+                    raise ValueError("metric evaluation does not exist")
```
kobicovaldev: [Required] This error message does not include `evaluation_id`. Would it be clearer to include it, like `f"metric evaluation {evaluation_id} does not exist"`?

runner/src/coval_bench/db/models.py
kobicovaldev: [Suggestion] Would it be cleaner to move `import math` to the module-level imports alongside `datetime` and `UUID`? It is lazily imported in four places (`_outcome_matches_error`, `_artifact_identity`, `_finite`, and `_valid_percentiles`), and the same pattern appears in `registries/metrics.py`. A single top-level import would be consistent and expose any missing dependency upfront.

runner/src/coval_bench/db/writer.py
kobicovaldev: [Required] This error message does not include `evaluation_id`, so it is hard to diagnose which evaluation hit this path. Would it be clearer to include it, like `f"metric evaluation {evaluation_id} is not queued"`?

runner/src/coval_bench/db/writer.py
kobicovaldev: [Required] This error message does not include `evaluation_id`. Would it be clearer to include it, like `f"metric evaluation {evaluation_id} is not queued or running"`?

runner/src/coval_bench/db/writer.py
kobicovaldev: [Required] This error message does not include `evaluation_id`. Would it be clearer to include it, like `f"metric evaluation {evaluation_id} does not exist"`?

kobicovaldev (changes requested):

runner/src/coval_bench/db/migrations/versions/20260813_0018_normalized_benchmark_storage.py
```diff
@@ -0,0 +1,206 @@
...
+            CHECK (updated_at >= created_at), CHECK (finished_at IS NULL OR (started_at IS NOT NULL AND finished_at >= started_at)),
+            CHECK ((status = 'queued' AND started_at IS NULL AND finished_at IS NULL AND error IS NULL) OR (status = 'running' AND started_at IS NOT NULL AND finished_at IS NULL AND error IS NULL) OR (status = 'partial' AND started_at IS NOT NULL AND finished_at IS NOT NULL) OR (status = 'succeeded' AND started_at IS NOT NULL AND finished_at IS NOT NULL AND error IS NULL) OR (status = 'failed' AND started_at IS NOT NULL AND finished_at IS NOT NULL AND error IS NOT NULL)),
+            UNIQUE (observation_id, metric_type, metric_version)
+        );
+        CREATE TABLE benchmarks_v2.metric_values (
+            metric_evaluation_id UUID NOT NULL REFERENCES benchmarks_v2.metric_evaluations(id) ON DELETE CASCADE,
+            value_key TEXT NOT NULL CHECK (value_key <> ''), unit TEXT NOT NULL CHECK (unit <> ''),
+            value DOUBLE PRECISION NOT NULL CHECK (value NOT IN ('NaN'::float8, 'Infinity'::float8, '-Infinity'::float8)),
+            is_primary BOOLEAN NOT NULL DEFAULT false, PRIMARY KEY (metric_evaluation_id, value_key)
```
kobicovaldev, line 88: [Suggestion] Have we considered avoiding a boolean database column here and modeling this as an enum or role field instead? Even though `is_primary` is tightly constrained today, boolean columns tend to become harder to query and evolve once a third state or another designation shows up.

callumreid (commented):
Staff review pass. Reviewed `341513da` against `origin/main`. All checks pass.

The review standard I apply treats a migration as a readiness blocker unless it is essential and safe. This one clears that bar comfortably and is unusually well built:

- purely additive, six new tables in `benchmarks_v2`, no ALTER or DROP against existing objects, and `runs`/`results`/`results_by_bucket` untouched;
- the `downgrade()` drops triggers, functions, indexes, and tables in correct dependency order, which most hand-written Alembic downgrades get wrong;
- no `CONCURRENTLY` needed because every index is on a table created empty in the same migration, so there is no lock or timeout exposure;
- the state machine is enforced in the database, not just in the writer: the `status` CHECK enumerates the transitions and pins which timestamp and error columns must be set in each state, and `metric_values_one_primary` is a partial unique index rather than an application invariant.

Three things worth settling before it applies.

### P2 - `UNIQUE NULLS NOT DISTINCT` needs PostgreSQL 15+, and it is load-bearing

```sql
UNIQUE NULLS NOT DISTINCT (run_id, dataset_id, sample_id, provider, model, voice, benchmark)
```

`NULLS NOT DISTINCT` arrived in PostgreSQL 15. It is not decoration here: `voice` is nullable, and standard NULL semantics would let unlimited duplicate observations coexist for every STT row where `voice IS NULL`, which is exactly the uniqueness this constraint exists to provide.

`docker-compose.yml` pins `postgres:16-alpine`, so local and CI are fine. The PR does not state the version of the deployed benchmarks database, and that is the one that matters. Worth confirming before apply, because on 14 or earlier the statement is a syntax error and the migration does not partially apply, it just fails, leaving the schema untouched and the deploy blocked.

### P2 - the accepted v1 limitation is lossy going forward, not merely deferred

> accepted v1 limitation: evaluations and preprocessing artifacts share an observation but do not store an exact input-link row

The uniqueness on `preprocessing_artifacts` is `(observation_id, pipeline, pipeline_version, artifact_name, producer_version)`, so a single observation can legitimately hold several artifacts differing only by `producer_version` or `pipeline_version`. With no link row, a `metric_evaluation` says which observation it belongs to and not which artifact it consumed.

The part that makes this more than scheduling: adding the link table in a later pass fixes rows written after that pass. Every evaluation written in the interim has provenance that cannot be reconstructed, because the information was never recorded anywhere. For a benchmarks system whose output is meant to be a reproducible measurement, "which aligner version produced the timestamps behind this WER number" is not a nice-to-have.

A cheap interim mitigation preserves the option: store the consumed artifact's `id` or `content_sha256` on `metric_evaluations` now, one nullable column, no join table, no design commitment. Later normalization can backfill from it. Worth deciding deliberately rather than accepting by default, since the cost of the decision is paid in data that will not exist.

### P3 - the state machine now has two authorities

The transition rules live in SQL CHECK constraints and functions, and again in the typed writer operations. Defense in depth is the right call for immutable measurement data and I would not remove either. The cost is that changing a status value or a legal transition requires a migration and a code change in lockstep, and the SQL half is the one that gets forgotten because it is invisible from the Python.

A comment on each side pointing at the other, naming the migration revision (`20260812_0017`) from the writer and the writer module from the migration docstring, makes that coupling discoverable at the moment someone edits one of them.

### P3 - the typed surface is the headline and is the least verified part

The verification section is admirably honest:

> full unit collection is unavailable in the current local environment because unrelated optional dependencies are absent (posthog, google-cloud-storage, whisper-normalizer, and pyarrow)
> mypy is not installed in the current coval environment

That leaves roughly 2000 lines of new writer and registry code, whose stated purpose is "typed writer operations and registered value contracts", with no type check locally, and it is not obvious from the check names whether Runner CI runs mypy. If it does, say so and this is a non-issue. If it does not, the contracts this PR is named for are currently unenforced by anything except the 34 focused tests.

### Not blocking

- Long-form `metric_values` rows with a singular unit and an explicit primary value is the right shape for a metric set that is expected to grow, and it is what lets the partial unique index express "exactly one primary" declaratively.
- Keeping non-scalar detail in private GCS artifacts behind typed pointers, with `gcs_uri` regex-constrained and `content_sha256` required, keeps the database narrow without losing addressability.
- The paired CHECK requiring all four audio columns to be present or all four absent is a good use of a table-level constraint; that quartet going half-populated is the classic ingestion bug.
- Not cutting over legacy storage in the same PR is the right sequencing.

**Verdict: NOT READY**, pending the PostgreSQL version confirmation and a decision on the provenance column. Neither is a redesign; the second is a one-column choice that gets much more expensive after the first rows land. Confidence: High on the migration reading; Medium on the deployed environment, which I could not inspect from the repository.

runner/src/coval_bench/db/migrations/versions/20260813_0018_normalized_benchmark_storage.py
```diff
@@ -0,0 +1,206 @@
...
+            status TEXT NOT NULL CHECK (status IN ('queued', 'running', 'partial', 'succeeded', 'failed')),
+            started_at TIMESTAMPTZ, finished_at TIMESTAMPTZ, error TEXT CHECK (error IS NULL OR error <> ''),
+            attempt_count INTEGER NOT NULL DEFAULT 0 CHECK (attempt_count >= 0),
+            created_at TIMESTAMPTZ NOT NULL DEFAULT now(), updated_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+            CHECK (updated_at >= created_at), CHECK (finished_at IS NULL OR (started_at IS NOT NULL AND finished_at >= started_at)),
+            CHECK ((status = 'queued' AND started_at IS NULL AND finished_at IS NULL AND error IS NULL) OR (status = 'running' AND started_at IS NOT NULL AND finished_at IS NULL AND error IS NULL) OR (status = 'partial' AND started_at IS NOT NULL AND finished_at IS NOT NULL) OR (status = 'succeeded' AND started_at IS NOT NULL AND finished_at IS NOT NULL AND error IS NULL) OR (status = 'failed' AND started_at IS NOT NULL AND finished_at IS NOT NULL AND error IS NOT NULL)),
+            UNIQUE (observation_id, metric_type, metric_version)
+        );
+        CREATE TABLE benchmarks_v2.metric_values (
```
kobicovaldev, line 84: [Suggestion] Have we considered drawing a stricter line between fields we know we need to query/index and fields that are still part of an evolving metric contract? The normalized shape is clean, but my concern is that metric storage is likely to keep changing as new LLM-assisted follow-up PRs propose small schema improvements. It might be safer to keep only durable identity/lifecycle/query fields in tables and put evolving metric-specific structure behind versioned contracts or JSON/artifact payloads until the model stabilizes.

kobicovaldev, line 84: [Suggestion] Would it be worth spelling out the API operations this storage layer needs to support before finalizing the table shape? For example: create observation, queue/start/complete/fail metric evaluation, attach artifacts, and read bucket rollups. Then we can derive the durable relational fields from the request/query patterns and keep evolving metric-specific structure behind versioned contracts or JSON/artifacts. That might reduce follow-up schema churn from future “cleaner” LLM-suggested refinements.

runner/src/coval_bench/db/migrations/versions/20260818_0018_normalized_benchmark_storage.py
```diff
@@ -0,0 +1,206 @@
...
+            status TEXT NOT NULL CHECK (status IN ('queued', 'running', 'partial', 'succeeded', 'failed')),
+            started_at TIMESTAMPTZ, finished_at TIMESTAMPTZ, error TEXT CHECK (error IS NULL OR error <> ''),
+            attempt_count INTEGER NOT NULL DEFAULT 0 CHECK (attempt_count >= 0),
+            created_at TIMESTAMPTZ NOT NULL DEFAULT now(), updated_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+            CHECK (updated_at >= created_at), CHECK (finished_at IS NULL OR (started_at IS NOT NULL AND finished_at >= started_at)),
+            CHECK ((status = 'queued' AND started_at IS NULL AND finished_at IS NULL AND error IS NULL) OR (status = 'running' AND started_at IS NOT NULL AND finished_at IS NULL AND error IS NULL) OR (status = 'partial' AND started_at IS NOT NULL AND finished_at IS NOT NULL) OR (status = 'succeeded' AND started_at IS NOT NULL AND finished_at IS NOT NULL AND error IS NULL) OR (status = 'failed' AND started_at IS NOT NULL AND finished_at IS NOT NULL AND error IS NOT NULL)),
+            UNIQUE (observation_id, metric_type, metric_version)
+        );
+        CREATE TABLE benchmarks_v2.metric_values (
```
kdmelon: adding it to infra repo

runner/src/coval_bench/db/models.py
```diff
@@ -32,9 +35,261 @@
...
+    "MetricEvaluationInput",
 ]
 
 
+class ProcessingStatus(StrEnum):
+    """Shared lifecycle status for normalized work rows."""
+
+    QUEUED = "queued"
+    RUNNING = "running"
+    PARTIAL = "partial"
```
kobicovaldev: Whats the plan around partial?

runner/src/coval_bench/db/models.py
```diff
@@ -32,9 +35,261 @@
...
+    SUCCEEDED = "succeeded"
+    FAILED = "failed"
+
+
+class ObservationStatus(StrEnum):
+    """Capture outcome of one benchmark sample/provider/model observation."""
+
+    SUCCEEDED = "succeeded"
+    FAILED = "failed"
```
kobicovaldev, line 71: Curious if you think error would be better than failed? Should we try to bake in the distinction of success, we failed, they failed from day one?

runner/src/coval_bench/db/models.py
```diff
@@ -32,9 +35,261 @@
...
+
+
+class ObservationSourceKind(StrEnum):
+    DATASET_AUDIO = "dataset_audio"
+    GENERATED_AUDIO = "generated_audio"
+    CONVERSATION_AUDIO = "conversation_audio"
```
kobicovaldev, line 79: This looks like its just covering the "types" of data sets but not enumerating the data sets. Is that correct? Just wanna be sure I understand.

runner/src/coval_bench/db/models.py
```diff
@@ -32,9 +35,261 @@
...
+class TimestampArtifactSchema(StrEnum):
+    """Canonical schemas emitted by timestamp preprocessing pipelines."""
+
+    WORD_TIMESTAMPS_V1 = "WordTimestampsV1"
+    PHONEME_TIMESTAMPS_V1 = "PhonemeTimestampsV1"
+
+
+class TimestampArtifactName(StrEnum):
+    WORD_TIMESTAMPS = "word_timestamps"
+    PHONEME_TIMESTAMPS = "phoneme_timestamps"
```
kobicovaldev, line 109: Are we planning to have filtering by these? Is that why we are including these at the top level? Should we use some jsonb fields?

runner/src/coval_bench/db/models.py
```diff
@@ -32,9 +35,261 @@
...
+        raise ValueError("must be a private gs:// object URI without query or fragment")
+    return value
+
+
+def _optional_private_gs_uri(value: str | None) -> str | None:
+    """Accept absent source audio while validating any supplied object URI."""
+    return None if value is None else _private_gs_uri(value)
+
+
+class Observation(BaseModel):
```
kobicovaldev, line 164: A handful of these seem like they should be in a jsonb. Do you agree?

runner/src/coval_bench/db/models.py
kdmelon: got ported from runs to show status for if some metrics were successful and some broke, will take out since it doesn't make sense for metrics

runner/src/coval_bench/db/models.py
kdmelon: will keep `status = failed` as the outcome and add `failure_origin = provider | runner` rather than renaming the state to error?

runner/src/coval_bench/db/models.py
kdmelon: ye, describes the source. `dataset_id` and `dataset_sha256` identify the actual dataset.

runner/src/coval_bench/db/writer.py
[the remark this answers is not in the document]
kobicovaldev: @/var/folders/2s/180s_7_j32bc6ws1qgrwg8yw0000gn/T/opencode/comment_period.md

runner/src/coval_bench/db/migrations/versions/20260813_0018_normalized_benchmark_storage.py
kobicovaldev: @/var/folders/2s/180s_7_j32bc6ws1qgrwg8yw0000gn/T/opencode/comment_isprimary.md

runner/src/coval_bench/db/migrations/versions/20260818_0018_normalized_benchmark_storage.py
kobicovaldev: @/var/folders/2s/180s_7_j32bc6ws1qgrwg8yw0000gn/T/opencode/comment_jsonb.md

runner/src/coval_bench/db/migrations/versions/20260813_0018_normalized_benchmark_storage.py
```diff
@@ -0,0 +1,242 @@
...
+            audio_uri TEXT CHECK (audio_uri IS NULL OR audio_uri ~ '^gs://[^/?#]+/[^/?#][^?#]*$'),
+            audio_sha256 TEXT CHECK (audio_sha256 IS NULL OR audio_sha256 ~ '^[0-9a-f]{64}$'),
+            audio_size_bytes BIGINT CHECK (audio_size_bytes IS NULL OR audio_size_bytes > 0),
+            audio_duration_ms INTEGER CHECK (audio_duration_ms IS NULL OR audio_duration_ms > 0),
+            captured_at TIMESTAMPTZ NOT NULL DEFAULT now(),
+            status TEXT NOT NULL CHECK (status IN ('succeeded', 'failed')),
+            error TEXT,
+            CHECK ((audio_uri IS NULL AND audio_sha256 IS NULL AND audio_size_bytes IS NULL AND audio_duration_ms IS NULL) OR (audio_uri IS NOT NULL AND audio_sha256 IS NOT NULL AND audio_size_bytes IS NOT NULL AND audio_duration_ms IS NOT NULL)),
+            CHECK ((status = 'succeeded' AND error IS NULL) OR (status = 'failed' AND error IS NOT NULL AND error <> '')),
+            UNIQUE NULLS NOT DISTINCT (run_id, dataset_id, sample_id, provider, model, voice, benchmark)
```
kobicovaldev: [Required] Before merge, please confirm the deployed benchmarks database is PostgreSQL 15 or newer. `NULLS NOT DISTINCT` requires PG15, and it is load-bearing here because `voice` is nullable: without it the observation uniqueness constraint silently stops working for every row where `voice IS NULL`. On PG14 or earlier this statement is a syntax error and the migration fails wholesale, which blocks the deploy. Local dev and CI pin postgres:16, so the only version that matters is the deployed one. Echoing callum's point but anchoring it to the line so it does not get lost.

runner/src/coval_bench/db/writer.py
```diff
@@ -154,6 +172,621 @@ async def record_results(self, results: Sequence[Result]) -> None:
...
+                    params,
+                )
+                await cur.execute(
+                    """
+                    INSERT INTO benchmarks_v2.metric_values_by_bucket
+                    (provider, model, benchmark, dataset_id, metric_type, metric_version,
+                     evaluation_variant, value_key,
+                     unit, bucket_at, min_value, p25, p50, p75, max_value, value_sum, sample_count)
+                    SELECT observation.provider, observation.model, observation.benchmark,
+                           COALESCE(observation.dataset_id, '__all__'), evaluation.metric_type,
```
ben-coval, line 760: Could you confirm the intended `__all__` behavior here? The second `GROUPING SET` omits `observation.dataset_id`, so PostgreSQL may require an explicit `GROUPING(observation.dataset_id)` expression around this `COALESCE` to produce the all-datasets row.

ben-coval (commented):
One inline question on the rollup grouping.

runner/src/coval_bench/db/models.py
kdmelon: yup, moved things around

runner/src/coval_bench/db/migrations/versions/20260813_0018_normalized_benchmark_storage.py
kdmelon: yes checked in an earlier comment

runner/src/coval_bench/db/models.py
kdmelon: yes, planning on using it for selecting the input for metrics and for supporting filtering and debugging

runner/src/coval_bench/db/writer.py
kdmelon: the bot said
> With PostgreSQL grouping sets, observation.dataset_id becomes NULL for the grouping set that omits it, and COALESCE(..., '__all__') correctly produces the all-datasets row. Since dataset_id is NOT NULL, there is no ambiguity requiring GROUPING(). The PostgreSQL test already verifies both per-dataset and __all__ rows.

kdmelon:
**TL;DR of changes from this review round**

- Removed the metric-level `partial` status.
- Added `failure_origin` to distinguish provider failures from runner failures.
- Clarified that `source_kind` is a category, while dataset fields identify the actual dataset.
- Made observation retries use the intended run/sample/provider/model/voice identity.
- Serialized metric-input creation against lifecycle transitions.
- Made preprocessing contracts forward-compatible without hardcoding every version into SQL.

**Observation flow**

- Added typed artifacts for transcripts, audio, traces, and timing events.
- Metrics can reference the exact raw or preprocessed artifacts they consumed.

**JSON boundary**

- Searchable and authoritative fields remain relational.
- JSONB is limited to optional, non-authoritative `provider_extras`.
- Large, evolving payloads remain versioned GCS artifacts.

kobicovaldev (approved):
Approving. The contract-to-registry refactor (moving metric and preprocessing contracts out of SQL) and the removal of the `partial` state address the main design concerns, and the new `observation_artifacts` table plus `provider_extras` jsonb handle the evolving capture metadata well. Two small items remain tracked in their threads as non-blocking follow-ups: the unused `period_seconds` parameter in `refresh_metric_values_bucket`, and the `is_primary` boolean question.

coval-cale (approved):

PR #500, opened 2026-08-13
Add interruption rate to the metric registry

seribaymadina:
Declares the S2S interruption rate metric — percent, lower is better — and the setting that holds its Coval metric id. Coval computes the metric itself, so ingesting it later needs no turn parsing.

Nothing reads either yet: this is the declaration the per-dataset metric scoping will name, split out so that change stays small. No migration, since results.metric_type is a plain text column.


cooperr1 (approved):
lgtm!

PR #502, opened 2026-08-13
Scope S2S metrics per caller condition

seribaymadina:
Each S2S dataset now declares which metric it requires, which it may carry, and which it never writes: the noise caller requires instruction adherence and never asks for V2V, while the standard caller still faults if latency goes missing.

That one declaration replaces the want_latency/want_instruction flag pairs, so a run is skipped once its declared metrics land instead of being rescanned until the ingest window closes.

The two row builders collapse into one, leaving interruption rate a small mapper and a dict entry away.


coval-cale (approved):
Hell ya

PR #506, opened 2026-08-14
Add ElevenLabs eleven_v3_conversational, retire the HTTP models

coval-cale:
Elevenlabs launched a websocket model!!!! Drop the http models and add that instead!


runner/src/coval_bench/providers/tts/elevenlabs.py
```diff
@@ -1,44 +1,38 @@
 # Copyright 2026 The Coval Benchmarks Authors
```
seribaymadina, line 1: nit: methodology.md still documents 11labs as an http tts provider

seribaymadina (approved):

PR #508, opened 2026-08-14
Scope S2S datasets to the test set that produced them

seribaymadina:
Groundwork only — every switch is off by default, so merging changes nothing in production.

- `AgentSpec` takes an optional test set and dataset family. Agents that set neither keep today's shared set and `s2s-multiturn-*` ids; ones that do file under `s2s-happypath-*` and can never share a row key with the shared board.
- `coval_s2s_condition_personas` replaces `coval_s2s_noisy_persona_id`. The old setting expressed "one persona is noisy, everything else is clean", and that fallback is how a new persona silently pools into the clean numbers. The map is exhaustive — unmapped personas now fail loudly, with an explicit `skip` for ones we deliberately don't ingest. Old setting still honoured while the map is unset, so the two can deploy in either order.
- Accented anchors on instruction adherence, not latency — those runs were scored without V2V attached, so anchoring on it would skip every one.
- Sampling publishes the clean caller only, and agents can opt out entirely. Two pre-launch models registered under codenames with `creator` unset.
- No migration; `dataset_id` has been a grouped column since the aggregates gained the dataset dimension.

Turning it on is a separate infra change, and the persona map should land on its own tick first so a missing persona is caught while only the four known-good agents are exposed to it.

### DIFF
--- a/runner/src/coval_bench/config.py
+++ b/runner/src/coval_bench/config.py
@@ -137,6 +137,8 @@ class Settings(BaseSettings):
     coval_s2s_gemini_agent_id: str | None = None
     coval_s2s_xai_agent_id: str | None = None
     coval_s2s_xai_think_fast_2_agent_id: str | None = None
+    coval_s2s_gray_agent_id: str | None = None
+    coval_s2s_red_agent_id: str | None = None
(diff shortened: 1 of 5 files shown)


coval-cale (approved):
Super cool dude!

PR #527, opened 2026-08-21
[BENCH-712] Move model lifecycle state to Postgres, add admin endpoints

coval-cale:
Implements the model state admin design (BENCH-710/BENCH-712): model lifecycle
state moves out of the registry into Postgres, toggleable via new coval-only
admin endpoints.

- `model_state` (running/shown booleans, natural key) + `model_state_history`,
  seeded in migration 20260820_0019 from a literal snapshot of today's statuses
  (ACTIVE→running+shown, PAUSED→shown, EARLY_ACCESS→running, RETIRED/PENDING→neither)
- `db/model_state.py`: the one read/write seam. Missing row → Hidden (new models
  run under embargo); orphan rows inert
- Rewired consumers: orchestrator selection, embargo set, `/v1/providers` flags,
  arena roster + leaderboard hiding. `/v1/providers` response schema unchanged
- Deleted `ModelStatus` and the registry `status` field
- `GET /v1/admin/models`, `PATCH /v1/admin/models/{benchmark}/{provider}/{model}`
  — require a verified `@coval.dev` Clerk email (active partner org never narrows it)
- ~10s state cache on the API; a PATCH busts the local instance immediately
- CORS: `PATCH` + `Content-Type` added

Behavior notes:
- `/v1/providers` now reads the DB and 503s without it (was "no DB hit") — fail
  closed rather than guess, since a fabricated default could leak a hidden model
- Offline tools (tune_scale, check_arena_keys) use `assume_all_active()`; the
  parity check now covers every arena-enabled provider, a superset of the live roster
- Two registry invariants weakened by the field removal: "active TTS models have
  voices/pools" is no longer statically checkable — replaced with pinned voiceless
  exceptions and structural pool checks; activating a voiceless model is now a
  runtime concern

Deploy: run `coval-bench db migrate` against prod before or with this release —
nothing in the deploy path runs it automatically.

<h3>Greptile Summary</h3>

The PR moves model scheduling and visibility state from registry constants into Postgres, adds authenticated administration endpoints and history, and rewires API, runner, and arena consumers.

- Adds persisted `running`/`shown` state with migration seeding and an audit-history table.
- Adds coval-only model administration endpoints and short-lived API state caching.
- Updates benchmark scheduling, embargo filtering, provider metadata, and arena roster/leaderboard behavior to consume database state.

The concurrent first-write audit defect should be fixed before merging so model lifecycle history remains trustworthy.

Missing model-state rows are supported, but concurrent updates can both observe no prior row and cause the later update to persist a false NULL prior state in the admin-visible history.

**Files Needing Attention:** runner/src/coval_bench/db/model_state.py

<sub>Reviews (1): Last reviewed commit: ["\[BENCH-712\] Move model lifecycle state t..."](https://github.com/coval-ai/benchmarks/commit/032c2f81e97fe8f1080d24fa05e0e42694c177f9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55622796)</sub>

<details><summary><h4>Context used (4)</h4></summary>

- Knowledge Base — [API service](https://app.greptile.com/coval/-/custom-context/knowledge-base/coval-ai/benchmarks/-/docs/api-service.md)
- Knowledge Base — [Arena subsystem](https://app.greptile.com/coval/-/custom-context/knowledge-base/coval-ai/benchmarks/-/docs/arena-subsystem.md)
</details>


runner/src/coval_bench/registries/models.py
```diff
@@ -97,18 +86,12 @@ class RegisteredModel(BaseModel, frozen=True, extra="forbid"):
     # unknown. Geo- or globally routed endpoints record where our runner's
     # traffic lands.
     region: Literal["us", "eu", "asia"] | None = None
-    status: ModelStatus
-    arena_enabled: bool = True  # in the arena roster? independent of dashboard `status`
+    arena_enabled: bool = True  # in the arena roster? independent of the dashboard state
```
callumreid, line 89: **P1 - Merging current `origin/main` makes the registry fail at import time.** A fresh merge retains the newly added `speechmatics/linden-1` entry with `status=_EARLY_ACCESS`, but this PR removes both `RegisteredModel.status` and `_EARLY_ACCESS`. Evaluation of that entry therefore raises `NameError` before the API or runner can start. Rebase and reconcile the entry, including its intended lifecycle state in the migration seed, and verify the merged registry imports.

runner/src/coval_bench/api/routers/admin_models.py
```diff
@@ -0,0 +1,137 @@
...
+                ModelStateChangeOut.model_validate(change, from_attributes=True)
+                for change in history.get((m.benchmark, m.provider, m.model), [])
+            ],
+        )
+        for m in MODEL_REGISTRY
+    ]
+    return AdminModelsResponse(models=models)
+
+
+@router.patch("/admin/models/{benchmark}/{provider}/{model}", response_model=ModelStateOut)
```
callumreid, line 96: **P2 - Models whose names contain `/` cannot be patched.** FastAPI's default string path parameter stops at a slash, so registered names such as `canopylabs/orpheus-v1-english` do not reach this handler and cannot be administered. Use a path converter or a stable encoded identifier, and add an API test that patches a slash-containing model name.

runner/src/coval_bench/db/model_state.py
```diff
@@ -0,0 +1,214 @@
...
+        "running": running,
+        "shown": shown,
+        "changed_by": changed_by,
+    }
+    async with (
+        pool.connection() as conn,
+        conn.transaction(),
+        conn.cursor(row_factory=psycopg.rows.dict_row) as cur,
+    ):
+        await cur.execute(
```
callumreid, line 127: **P2 - Concurrent first writes produce incorrect audit history.** `SELECT ... FOR UPDATE` cannot lock an absent row, so two transactions can both observe `old is None`; after the upserts serialize on the unique key, both history records still claim a null prior state. Serialize on the natural key before reading (for example with a transaction advisory lock), and add a concurrent first-write test proving the second audit row records the first write.

runner/src/coval_bench/api/routers/admin_models.py
```diff
@@ -0,0 +1,137 @@
...
+        provider=provider,
+        model=model,
+        running=body.running,
+        shown=body.shown,
+        changed_by=email,
+    )
+    # Bust this instance's state cache so the toggle is immediately visible to
+    # every endpoint, not just after the TTL. Other instances converge within
+    # the TTL on their own.
+    request.app.state.model_state_cache.pop(MODEL_STATE_CACHE_KEY, None)
```
callumreid, line 127: **P2 - Invalidation can lose a race with an in-flight cache fill.** If another request misses the cache and starts the database fetch before this `pop`, it can publish its stale result after the mutation and repopulate the full TTL. Perform the mutation/invalidation while holding the same per-key lock used by cache fills, so no pre-write fill can land after invalidation.

runner/src/coval_bench/api/routers/admin_models.py
```diff
@@ -0,0 +1,137 @@
...
+    benchmark: Benchmark,
+    provider: str,
+    model: str,
+    body: ModelStatePatch,
+    pool: AsyncConnectionPool[Any] = Depends(get_pool),
+    email: str = Depends(require_internal),
+) -> ModelStateOut:
+    """Set one model's state; the change lands in the history table."""
+    never_shared(response)
+    if not any(
```
callumreid, line 110: **P2 - The admin API can enable TTS entries that the runner cannot execute.** The registry intentionally retains voiceless TTS entries such as `openai/gpt-realtime-2025-08-28` and `cartesia/sonic`; this existence-only check permits `running=true`, after which orchestration has no voice to synthesize. Reject that state transition or encode run capability in the registry, with an API regression test.

runner/scripts/check_arena_keys.py
```diff
@@ -86,7 +87,9 @@ def main(tf_path: str) -> int:
 
     required: set[str] = set()
     unmapped: set[str] = set()
-    for m in active_tts_models():
+    # Run-state lives in the DB and can flip without a deploy, so the check
+    # covers every arena-enabled provider, not just the currently live roster.
+    for m in active_tts_models(assume_all_active()):
```
callumreid, line 92: **P2 - Exact-head CI fails because this expands eligibility to an unmounted key.** `assume_all_active()` makes the Baseten TTS entry require `BASETEN_API_KEY`, but benchmarks-api does not mount it, so the `Arena provider keys mounted in infra` check is red at this SHA. Add the intended mount or exclude providers that are not actually arena-eligible, then make the parity check pass.

callumreid (changes requested):
**Verdict: NOT READY.** Reviewed exact head `032c2f81e97fe8f1080d24fa05e0e42694c177f9` against freshly fetched `origin/main`. A fresh merge simulation retains the new `speechmatics/linden-1` legacy status reference after this head removes that API, and exact-head CI is red on `Arena provider keys mounted in infra` while the remaining reported checks pass. The inline findings are definite blockers.

coval-cale:
Closing. Synced with Ben and going with a slightly different architecture/implementation plan

PR #521, opened 2026-08-19
[BENCH-704] Retire the X-Internal-Key and X-EA-Token proofs

coval-cale:
The Clerk bearer session token becomes the single early-access proof, per BENCH-704. With org grants able to name whole providers or individual `provider/model` pairs (and exclusive orgs, #519), the `X-Internal-Key` and `X-EA-Token` header paths were redundant surface, so this removes them.

- `hidden_early_access` resolves visibility from the `Authorization` bearer alone; `is_internal`, the token-allowlist parsing, and the `INTERNAL_API_KEY` / `EARLY_ACCESS_TOKENS` settings are gone.
- `Vary` shrinks to `Authorization`, and CORS stops allowing the retired headers.
- The `X-EA-Token-Status` response header keeps its name and its `absent` / `unknown` / `accepted` values, but `internal` disappears — a coval.dev session reads as `accepted`.
- Tests mint sessions against a stubbed Clerk instance in the API conftest; the end-to-end embargo suite now proves the same properties (cache isolation between two org grants, per-model scoping, no-store on privileged responses) over bearer proofs, plus a regression test that the retired headers are ignored. `test_early_access_scope.py` is deleted with its subject; its retired-board-key coverage moved to `test_clerk_scope.py`.

Ops note before deploy: any partner still presenting a raw `X-EA-Token` needs a Clerk org with a matching `CLERK_ORG_PROVIDERS` entry, and anything internal presenting `X-Internal-Key` needs a coval.dev session token instead.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Greptile Summary</h3>

The PR retires the X-Internal-Key and X-EA-Token early-access proofs, leaving Clerk bearer sessions as the sole authorization mechanism.

- Removes the legacy proof settings, parsing, CORS allowances, and cache variation headers.
- Preserves additive and exclusive organization grants, retired model-key handling, and caller-specific cache isolation.
- Reworks API fixtures and embargo tests to use locally signed Clerk sessions and verifies that retired headers are ignored.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge with no actionable correctness or security defects identified.

Early-access filtering remains consistently derived from verified Clerk sessions, exclusive grants still take precedence, and privileged responses retain caller-specific cache protections.

[...]

PR #537, opened 2026-08-24
Assert the retired proof headers stay off the CORS allowlist

seribaymadina:
A browser only sends a request whose headers the preflight approved, so putting `X-EA-Token` or `X-Internal-Key` back on the allowlist is not a narrow early-access regression: any client still sending one has its request cancelled before it leaves the browser, every leaderboard renders empty including the public rows, and nothing reaches a server to log it. That is how the last one went unnoticed.

#521 left the positive case for `Authorization`; this is its negative twin. Test-only, and it can fail only if CORS is re-widened.

Verified by mutation: re-adding both headers to `allow_headers` fails both cases, reverting passes. Fixes the client side in coval-ai/benchmarks-web#47.

<h3>Greptile Summary</h3>

The PR adds regression coverage ensuring the retired `X-Internal-Key` and `X-EA-Token` headers remain excluded from CORS preflight approval.
- Parameterizes both retired proof headers.
- Verifies each preflight is rejected and omitted from `Access-Control-Allow-Headers`.

The PR appears safe to merge.

No blocking failure remains.

<sub>Reviews (2): Last reviewed commit: ["Assert the retired proof headers stay of..."](https://github.com/coval-ai/benchmarks/commit/d4600aa1ae2c2630b06865d630241e8255166458) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56416877)</sub>

### DIFF
--- a/runner/tests/api/test_cors.py
+++ b/runner/tests/api/test_cors.py
@@ -5,7 +5,6 @@
 
 from __future__ import annotations
 
-import pytest
 from httpx import AsyncClient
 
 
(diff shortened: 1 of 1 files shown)


coval-cale (commented):
Does this involve network calls or is this locally mocked? Cause if it's network calls we should refactor this test suite to not use them.

seribaymadina:
Locally mocked — no network. The `client` fixture uses `ASGITransport(app=app)` against `base_url="http://test"` (`tests/api/conftest.py:346`), so requests go straight into the ASGI app in-process; conftest's module docstring says the same on line 12. Nothing to refactor here.

Separately, you prompted a rethink of the test itself and I've replaced it in `8f4936e`. The parametrized negative only caught headers being *added* to the allowlist — but this outage came from one being *removed*: #521 dropped `X-EA-Token`/`X-Internal-Key` while the dashboard still sent them, and my test would have passed happily through that. It was guarding the wrong direction.

It now asserts the allowlist is exactly `{accept, accept-language, authorization, content-language, content-type}`, so either direction fails. Mutation-checked both ways: adding `X-EA-Token` fails it, and dropping `Authorization` (the #521 shape) fails it too.

Worth being blunt about the limit: this pins the server's side only. It can't see what the dashboard sends, so it turns a silent narrowing into a failing test that asks "what still sends this?" rather than actually verifying the contract. The real guard is a cross-repo check that the client's header set is a subset of this one — happy to file that separately if it's worth it.

Client-side fix is coval-ai/benchmarks-web#47.

coval-cale (approved):

PR #136, opened 2026-06-17
Mark methodology changes on benchmark timeline charts

seribaymadina:
Closes BENCH-301.

When a timeline steps up or down, you can't tell if a provider changed or if we changed how we measure. This adds a marker that explains the step.

- A dashed amber line with a **(!)** badge at each methodology change; hover shows the title, date, and detail.
- Markers only show on the metric they affect, and only when the change falls in the visible time range.
- Changes are curated in `web/lib/config/methodologyChanges.ts`. Seeded with the TTFS end-of-speech anchor, Grok TTFT exclusion, and Gradium event-driven finalization (dates from PR merges — worth confirming).
- Adds an advisory CI check that warns when metric or dataset code changes without a matching marker entry. Provider add/exclude is left to human judgment.

No backend changes — markers read a static config file.

## Summary by CodeRabbit

* **New Features**
  * Timeline charts now display methodology change markers that highlight significant updates to metric definitions. Hovering over any marker reveals the change title, date, and detailed explanation of the impact.

<h3>Greptile Summary</h3>

Adds amber dashed `ReferenceLine` markers to the timeline charts that appear at the exact x-position of each methodology change, filtered to the active metric and visible time window. Hovering the badge shows a popover with the change title, date, and detail text. A new advisory CI job warns (but never blocks) when metric or dataset files change without a matching entry in the new `methodologyChanges.ts` config.

- `web/lib/config/methodologyChanges.ts` — static registry of 4 methodology events scoped to `ttfa` and `ttfs`; the previous UTC-midnight timezone bug is fixed with `T12:00:00Z` anchoring and the marker list is correctly wrapped in `useMemo`.
- `web/components/visualizations/TimelineChart.tsx` — `MethodologyMarkerLabel` SVG component with an enlarged transparent hit rect and hover callbacks; popover state (`hoveredMarker`) is cleared on metric or time-window change.
- `.github/workflows/ci.yml` — `methodology-gate` job watches metric files (including `ttfs`) and `docs/methodology.md`, emitting a `::warning::` when those paths change without a matching config update.

<h3>Confidence Score: 5/5</h3>

Safe to merge — all previous review findings are addressed and no new defects were introduced.

[...]

PR #543, opened 2026-08-25
Give the dental appointment runs their own S2S family

seribaymadina:
The gray and red agents moved to the dental appointment test set on 08-20, but the fetch still filtered on the happy-path one. `_list_runs` returned zero rows, so `newest_data_at` stayed None and both providers went `provider_stale` → `failed` on every tick — no exception, exit 0. Nothing has ingested for them since.

Repointing `coval_s2s_happypath_test_set_id` would fix the symptom and break the data: dental is a different domain with 50 cases against happy-path's 30, and filing both under `s2s-happypath-*` pools two populations under one dataset id, which is the anchor every optional metric matches against. So dental gets its own family, dataset ids, and setting.

Clean anchors on V2V; noisy and accented anchor on instruction following. Noisy because background noise sits in the lane the turn boundaries come from, accented to match its happy-path sibling — the same persona's runs are scored without the latency metric there, and anchoring on a metric the runs lack would skip every one and leave the condition silently empty.

Also makes a missing test set fail loudly once one of the family's agents is configured. Unset previously skipped the agent with a warning indistinguishable from never having configured it, which is the exact failure mode above.

Paired with coval-ai/benchmark-infra#136, which supplies `COVAL_S2S_DENTAL_TEST_SET_ID`. **Apply #136 first, then ship this.** `extra="ignore"` means the deployed image ignores the unknown var, so applying early is a no-op; shipping this image first with the var unset leaves both agents producing no rows until the apply lands.

Happy-path is deliberately left in place — no `AgentSpec` references it after this change, but the family, its dataset ids, and its setting stay for now.

### DIFF
--- a/runner/src/coval_bench/s2s/conditions.py
+++ b/runner/src/coval_bench/s2s/conditions.py
@@ -153,9 +153,12 @@ CONDITIONS: dict[str, DatasetMetrics] = {
         required=Metric.INSTRUCTION_FOLLOWING,
         optional=frozenset({Metric.INTERRUPTION_RATE}),
     ),
+    # Anchored like its happy-path sibling: the same accented persona's runs are
+    # scored without V2V there, and anchoring on a metric the runs lack would
+    # skip every one of them and leave the condition silently empty.
     DATASET_ID_DENTAL_ACCENTED: DatasetMetrics(
-        required=Metric.V2V,
-        optional=frozenset({Metric.INSTRUCTION_FOLLOWING, Metric.INTERRUPTION_RATE}),
+        required=Metric.INSTRUCTION_FOLLOWING,
+        optional=frozenset({Metric.INTERRUPTION_RATE}),
     ),
 }
 
--- a/runner/src/coval_bench/s2s/conditions.py
+++ b/runner/src/coval_bench/s2s/conditions.py
@@ -21,12 +21,16 @@ from coval_bench.registries import Metric
 
 __all__ = [
     "DATASET_ID",
+    "DATASET_ID_DENTAL",
+    "DATASET_ID_DENTAL_ACCENTED",
+    "DATASET_ID_DENTAL_NOISY",
     "DATASET_ID_HAPPYPATH",
     "DATASET_ID_HAPPYPATH_ACCENTED",
     "DATASET_ID_HAPPYPATH_NOISY",
     "DATASET_ID_MULTITURN",
     "DATASET_ID_MULTITURN_NOISY",
     "DEFAULT_CONDITION",
+    "FAMILY_DENTAL",
     "FAMILY_HAPPYPATH",
     "FAMILY_MULTITURN",
     "Condition",
@@ -56,6 +60,10 @@ class Condition(StrEnum):
 # same 30 conversations either way.
 FAMILY_MULTITURN = "s2s-multiturn"  # customer-service, the shared board
 FAMILY_HAPPYPATH = "s2s-happypath"
+# Dental appointment booking, its own domain and its own 50 cases. Separate from
+# happy-path on purpose: pooling two populations under one dataset id would break
+# the one-dataset-id = one-condition = one-population anchor the metrics rest on.
+FAMILY_DENTAL = "s2s-dental"
 
 # Single-turn SLURP manifest (legacy, latency only) and the multi-turn Coval test
 # set, split by caller condition so background noise never pools into the clean
@@ -68,6 +76,10 @@ DATASET_ID_HAPPYPATH = "s2s-happypath-v1"
 DATASET_ID_HAPPYPATH_NOISY = "s2s-happypath-noisy-v1"
 DATASET_ID_HAPPYPATH_ACCENTED = "s2s-happypath-accented-v1"
 
+DATASET_ID_DENTAL = "s2s-dental-v1"
+DATASET_ID_DENTAL_NOISY = "s2s-dental-noisy-v1"
+DATASET_ID_DENTAL_ACCENTED = "s2s-dental-accented-v1"
+
 # Unlisted pairs are a configuration error, not a silent skip.
 DATASET_IDS: dict[tuple[str, Condition], str] = {
     (FAMILY_MULTITURN, Condition.CLEAN): DATASET_ID_MULTITURN,
@@ -75,6 +87,9 @@ DATASET_IDS: dict[tuple[str, Condition], str] = {
     (FAMILY_HAPPYPATH, Condition.CLEAN): DATASET_ID_HAPPYPATH,
     (FAMILY_HAPPYPATH, Condition.NOISY): DATASET_ID_HAPPYPATH_NOISY,
     (FAMILY_HAPPYPATH, Condition.ACCENTED): DATASET_ID_HAPPYPATH_ACCENTED,
+    (FAMILY_DENTAL, Condition.CLEAN): DATASET_ID_DENTAL,
+    (FAMILY_DENTAL, Condition.NOISY): DATASET_ID_DENTAL_NOISY,
+    (FAMILY_DENTAL, Condition.ACCENTED): DATASET_ID_DENTAL_ACCENTED,
 }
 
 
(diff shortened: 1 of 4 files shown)


coval-cale (approved):

runner/src/coval_bench/s2s/conditions.py
```diff
@@ -130,6 +145,18 @@ def fetched(self) -> frozenset[Metric]:
...
+        required=Metric.V2V,
+        optional=frozenset({Metric.INSTRUCTION_FOLLOWING, Metric.INTERRUPTION_RATE}),
+    ),
+    DATASET_ID_DENTAL_NOISY: DatasetMetrics(
+        required=Metric.INSTRUCTION_FOLLOWING,
+        optional=frozenset({Metric.INTERRUPTION_RATE}),
+    ),
+    DATASET_ID_DENTAL_ACCENTED: DatasetMetrics(
+        required=Metric.V2V,
+        optional=frozenset({Metric.INSTRUCTION_FOLLOWING, Metric.INTERRUPTION_RATE}),
```
callumreid: **P2 — this accented anchor contradicts its happy-path sibling; if it is the wrong choice the whole condition goes silently empty.**

Twelve lines above, `DATASET_ID_HAPPYPATH_ACCENTED` anchors on `INSTRUCTION_FOLLOWING` with an explicit reason: *"These runs were scored without V2V attached, so anchoring on it would skip every one with `required_metric_absent`."* That entry covers the **same two agents** (gray/red) and the **same accented persona** — `EgwG6KHFcnaxEWqeZQPrm5` comes from the shared `COVAL_S2S_CONDITION_PERSONAS` map, which applies to every agent regardless of family. This entry anchors on `V2V` instead.

If the dental accented runs also arrive without the latency metric, every one of them hits `fetch_v2v.py:492-505`, logs `required_metric_absent`, and returns `None` before any DB write. Nothing else catches it: the clean personas keep ingesting so `data_seen` is set and the staleness check passes, and the provider still reports SUCCEEDED/PARTIAL. `s2s-dental-accented-v1` would just stay permanently empty, with warning-level log lines as the only evidence.

**I could not verify this either way** — whether V2V is attached to the accented sims on test set `7rJxSKMp` is Coval-side config I have no access to. If you have already confirmed those runs carry it, disregard this. If not, mirroring the happy-path entry (`required=Metric.INSTRUCTION_FOLLOWING`, `optional={INTERRUPTION_RATE}`) fails safe — and the note there already documents moving the anchor back once runs are re-scored with V2V.

runner/src/coval_bench/s2s/fetch_v2v.py
```diff
@@ -96,16 +96,16 @@ class CovalRun:
         agent_id_attr="coval_s2s_gray_agent_id",
         provider="colors",
         model="gray",
-        test_set_id_attr="coval_s2s_happypath_test_set_id",
-        family=FAMILY_HAPPYPATH,
+        test_set_id_attr="coval_s2s_dental_test_set_id",
```
callumreid, line 99: **P2 — this retires the happy-path family outright; worth making that an explicit decision rather than a side effect.**

gray and red were the *only* two `AgentSpec`s carrying `family=FAMILY_HAPPYPATH` — I checked `origin/main`, where they sit at lines 100 and 108. After this change no `AgentSpec` references it, so the following become unreachable at runtime: `FAMILY_HAPPYPATH` and `DATASET_ID_HAPPYPATH{,_NOISY,_ACCENTED}` with their three `CONDITIONS` entries; `coval_s2s_happypath_test_set_id` and its blank guard at `fetch_v2v.py:900-902`; and benchmark-infra's `COVAL_S2S_HAPPYPATH_TEST_SET_ID = "XAvmHX2o"`, which #136 deliberately leaves in place. No `s2s-happypath-*` row will ever be written again.

Two consequences worth deciding on deliberately:

1. **The existing series cannot be migrated.** Ingest dedupes on `(provider, coval_run_id, metric_type)` via `writer.coval_metric_ingested`, and `dataset_id` is not part of that key. Any gray/red run already ingested under `s2s-happypath-*` returns empty `pending` and is skipped as `run_already_ingested`, so a targeted backfill cannot re-key it. Whatever history exists splits permanently at the cutover.
2. **The dead config becomes a trap.** The `config.py` comment still reads "Test set for agents evaluated on the happy-path scenarios", but there will be no such agents. If happy-path is genuinely retired, dropping the setting and the terraform entry in the same pair keeps the two repos honest; if a happy-path agent is coming back, a one-line note here would say so.

Verified not affected: the public leaderboard is pinned elsewhere (`api/routers/leaderboard.py:58`, `_PRIMARY_DATASET_BY_BENCHMARK = {"S2S": "s2s-multiturn-v1"}`), and nothing in `docs/` references the happy-path family.

callumreid (approved):
**APPROVE** — the family wiring is complete and internally consistent, and the cross-repo identifier contract with coval-ai/benchmark-infra#136 matches exactly. Two P2s inline, both about consequences that are invisible from this diff alone.

**Reviewed:** merge base `076469a9d71b4756dff5878d87677ee1ddb19695` (three-dot; `main` is 1 commit ahead — #542, model-registry seeding, disjoint file set). Head `d7e0ce52f49df133fb8dc31acac3ba4445158bd4`. Observed 4 files / +51/-5, matching GitHub.

**Verified clean:**

- `test_set_id_attr="coval_s2s_dental_test_set_id"` matches the declared field character-for-character (`config.py:171`). This one is self-enforcing — `getattr(settings, spec.test_set_id_attr)` at `fetch_v2v.py:948` would `AttributeError` out of the entire fetch on a typo — but I checked it against the terraform side too, and `COVAL_S2S_DENTAL_TEST_SET_ID` resolves through `case_sensitive=False`.
- Family/dataset wiring is complete: all three `(FAMILY_DENTAL, Condition.*)` pairs are in `DATASET_IDS`, so `dataset_id_for` never raises for a dental persona. `Condition.SKIP` still short-circuits first.
- `DATASET_ID_DENTAL_NOISY` correctly drops V2V and anchors on `INSTRUCTION_FOLLOWING`, matching both noisy siblings and the "noise sits in the lane Silero VAD uses for turn boundaries" rationale. `DATASET_ID_DENTAL` matches `DATASET_ID_HAPPYPATH` exactly.
- The three new dataset-id strings collide with nothing else in the repo; `__all__` updated for all four names and kept sorted.
- The new blank guard mirrors the happy-path one exactly (rejects whitespace-only, allows `None`) and is covered by a test that would fail without it.
- `_expected_sample_models` still excludes gray/red (`publish_samples=False`), so no new `samples_provider_missing` alerts — the existing assertion at `test_s2s_fetch.py:285-295` is undisturbed.
- No migration, no schema change, no secret, no legacy-API usage.

**Quality gates: green on the exact head** `d7e0ce52f49d` — all five checks carry that `head_sha`. I confirmed `Runner CI` actually ran rather than skipping: the log shows `ruff check .` → "All checks passed!", `ruff format --check`, and `pytest -q --disable-socket --cov=coval_bench`. **CodeRabbit's green tick is not review coverage** — the org is on the Free plan, so it produces a summary and walkthrough only; its one pre-merge flag ("Docstring Coverage 25%") is noise for four config constants. No human review yet.

**PR description — worth a rewrite before merge.** The body is the four-word stub "add dental appointment runs". It omits the actual behavioural change (gray and red are being repointed *off* the happy-path family) and the cross-repo dependency. #136 cross-references this PR; this PR does not cross-reference #136, so a reader landing here alone has no way to know an infra apply is required for any of it to take effect.

**Ordering, which neither PR body states correctly:** apply **benchmark-infra#136 first**, then ship this. #136's body says the opposite. Applying #136 early is a no-op — `extra="ignore"` in `Settings.model_config` means the deployed image ignores the unknown var, and gray/red keep reading the happy-path id this PR still leaves in `config.py`. Shipping this image first with the var unset means both agents log `test_set_unset` and `continue`, producing no rows and no provider status until the apply lands, with nothing louder than a warning line to show for it. Self-healing inside the scan window, hence P2 — but there is no reason to take it. Note #136's apply is currently blocked on an unrelated `db_grants.tf` IAM failure inherited from its own `main`, so that ordering may need active sequencing rather than assuming.

**Residual risk:** I could not verify Coval-side facts — that test set `7rJxSKMp` exists, that it holds the 50 cases the comment claims, or whether its accented runs carry the V2V metric (that last one is the first P2 below). No benchmarks-org API key is available locally.

PR #545, opened 2026-08-25
Cale/bench 712 read the roster from postgres

coval-cale:
Lot of changed files but its basically pulling in the models from the new model endpoint instead of a registry.

Added since the first review:
- Merge of main plus migration `20260831_0023`: syncs the model tables with the six registry commits that landed after the seed (5 new models, 3 state flips). The parity test now compares by natural key — post-seed models take ids at the end, so id order diverges from literal order.
- The collected/published swap: `ModelStatus` is gone; `RegisteredModel`, every consumer, and the registry literals carry the two booleans the DB already stores. Migration `20260831_0024` publishes the 21 launched-then-retired models so they stay on the public catalogue; never-launched models stay unpublished.
- Behaviour shift that falls out of the swap: paused models now come back `disabled=true` from /v1/providers (they grey out and drop from rankings), and never-launched retired/pending models are omitted for public callers instead of listed disabled.


seribaymadina (approved):
seems non breaking to me!

PR #546, opened 2026-08-25
feat(contracts): add pinned stack and pull-contract command

seribaymadina:
Scaffolding for the voice orchestration benchmark, which ranks orchestration
platforms (Vapi, Telnyx, Twilio, LiveKit, Pipecat) rather than models. Design doc:
`internal-docs/design-docs/voice-orchestration-benchmark.md`.

**Nothing runs a benchmark yet.** This lands the contract variants share and the
command that extracts it. 7 lines of existing code touched: one import, one
`add_command`, two optional Settings fields. No behaviour changes.

| | |
|---|---|
| `contracts/stack.json` | the pinned component layer |
| `contracts/__init__.py` | `Stack` model, `load_stack`, `contract_sha256`, `stack_as_dict` |
| `contracts/dental/` | the suite |
| `variants/platforms.py` | per-platform fetcher table |
| `variants/pull.py` | `coval-bench pull-contract` |
| `tests/test_contracts.py` | 7 tests |

**1. Is the public/private line right?** Public: `system-prompt.txt`,
`first-message.txt`, `tool-definitions.json`. A reader cannot audit a result
without them. Private (`dental/_private/`, gitignored as a directory): scenario
scripts, assertions, and `mock-tools.json`, which encodes expected availability
and patient identities.

**2. The hash covers private files, so a fresh checkout hashes differently.**
Deliberate: a published number identifies which fixtures ran without revealing
them. `has_private_contract()` lets callers skip. Object if you want a
reproducible-anywhere hash instead.

**3. Does `platforms.py` fit the other four?** One entry per platform: API base,
key env var, where the prompt and tools sit in the response. Vapi is first
because it is the first variant standing. If Telnyx or Pipecat needs a fourth
field, better to know now.

**4. Unknown keys are rejected, `_`-prefixed ones allowed.** Catches
`{"modle": "gpt-4.1"}` instead of silently dropping the pin and leaving the
platform on its default. `stack_as_dict()` strips them before any vendor call.

**5. The pins.** `gpt-4.1` temp 0, `deepgram/nova-3`, ElevenLabs
`eleven_flash_v2`, `PCMU` 8 kHz, 800 ms end-of-turn, native auto-hangup off,
vendor post-call analysis off. Transport is per variant, not pinned. Hash
`146ac237d7ef007c`, pinned in a test.

### DIFF
--- a/runner/src/coval_bench/contracts/README.md
+++ b/runner/src/coval_bench/contracts/README.md
@@ -1,0 +1,121 @@
+# Contracts
+
+What every variant of the voice orchestration benchmark is given, and the one
+place a pinned value is written down.
+
+## Why this lives inside the package
+
+`Dockerfile.api` builds from `runner/` and copies only `pyproject.toml`,
+`uv.lock`, `README.md`, `src/` and `alembic.ini`. The mock-tool router runs
+inside that container and reads `tool-definitions.json` and the fixtures at
+request time, so a contract at the repo root would not exist when it is needed.
+Everything here ships in the wheel and is read with `importlib.resources`, the
+same way `datasets/manifests` is.
+
+## Layout
+
+```
+contracts/
+  stack.json          the pinned component layer, identical for every variant
+  dental/
+    system-prompt.txt      \  the agent under test. byte-identical across
+    first-message.txt       |  variants, covered by the contract hash
+    tool-definitions.json  /
+    _source/               pulled from the platform, never edited by hand
+    _private/              the evaluator. hashed, never committed
+  platforms/
+    <platform>-agent.json  the only permitted difference, published for audit
+```
+
+## The two rules
+
+1. **`stack.json` and the public `dental/` files are byte-identical across every variant.** A
+   platform config may not restate a pinned value; it references this. If two
+   platforms could drift apart, the comparison is void.
+2. **`platforms/` holds the declared difference.** Each file records how one
+   vendor was wired, so a reader can audit the tested configuration rather than
+   infer it from results.
+
+Transport is in neither. It is declared per variant on the registry row, because
+it is the one layer not pinned to a single value.
+
+## Rationale keys
+
+JSON has no comments, so rationale sits beside the value it explains under keys
+starting with `_`:
+
+```json
+"llm": {
+  "provider": "openai",
+  "model": "gpt-4.1",
+  "temperature": 0,
+  "_why": "Pinnable on every platform under test, so model choice cannot explain a difference."
+}
+```
+
+Models accept `_`-prefixed extras and reject anything else, so a mistyped key
+still fails loudly:
+
+```
+{"modle": "gpt-4.1"}  ->  ValidationError: field 'model' missing
+```
+
+That is the failure worth catching. A mistyped *value* is caught later, when the
+vendor rejects it at apply time. A mistyped *key* would silently drop the pin and
+leave the platform on its own default, which is invisible in the results.
+
+`stack_as_dict()` strips these before anything is sent to a vendor.
+
+## The contract hash
+
+`contract_sha256("dental")` is one SHA-256 over `stack.json` plus the suite
+files, in a fixed order. It is published beside every result, so a number always
+points at an exact contract. It covers `stack.json` because changing a pin
+changes what the agent is just as surely as changing its prompt.
+
+Missing files are skipped, so the hash is meaningful before the contract is
+complete and changes when a file is added.
+
+## `_source/`
+
+Raw dumps from `coval-bench pull-contract`. Committed on purpose: re-running the
+pull produces a diff, and a diff means a vendor changed something underneath us.
+That is drift detection for free.
+
+Promoting a `_source/` file into a contract file is a deliberate human step, so a
+re-pull can never silently rewrite a contract that a published run already used.
+
+### What is published, and what is not
+
+The agent definition is public; the evaluator is not.
+
+| File | Published | Why |
+|---|---|---|
+| `system-prompt.txt`, `first-message.txt`, `tool-definitions.json` | yes | The agent under test. A reader must be able to audit what the agent was told in order to trust a result |
+| `_source/coval-agent.json`, `coval-prompt.txt`, `test-set.json` | yes, identifiers withheld | Provenance and coverage-level methodology. No scenario specifics: no patient names, dates or availability |
+| `_private/` | **no, gitignored as a directory** | The evaluator. `input_str` is the caller's script, `expected_behaviors` are the assertions, and `mock-tools.json` encodes the expected availability and patient identities. Publishing them would let a platform optimise for the test rather than the task |
+
+Identifiers are withheld because this repo is public. `redact_identifiers()`
+replaces `id`, `name`, `customer_agent_id`, `phone_number` and `endpoint` with
+`[REDACTED]` as the dump is written, so a re-pull cannot reintroduce them. The
+dial target is the reason it matters: it is a live endpoint, and publishing it
+lets anyone call the assistant. `display_name` and `voice_id` are deliberately
+kept, being methodology rather than identity.
+
+The contract hash is unaffected: `contract_sha256` covers `stack.json` and the suite files, never `_source/`.
+
+Two things to know about what is in there:
+
+- `coval-prompt.txt` is Coval's `prompt` field, which for a `MODEL_TYPE_VOICE`
+  agent holds the **benchmark specification**, not the prompt the agent runs. The
+  running prompt lives with the vendor and arrives as `vapi-prompt.txt`.
+- The scan before committing found no credential-shaped fields, all emails on
+  `@example.com`, and every phone number in the 555 block except
+  `602-556-0182`, which is the deliberate confusable for IR03 (phone digit
+  repair) and is fictional.
+
+Platform configs are redacted on write: any key matching
+`api_key|secret|token|password|bearer|authorization` becomes `[REDACTED]` before
+the file is created, and the puller prints what it redacted. `credentialIds` is
+deliberately not matched, being an account-scoped reference that `apply` needs
+and that carries no key material.
--- a/runner/src/coval_bench/variants/pull.py
+++ b/runner/src/coval_bench/variants/pull.py
@@ -1,0 +1,239 @@
+# Copyright 2026 The Coval Benchmarks Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Pull a live suite contract out of Coval and, optionally, one platform.
+
+The contract is *extracted*, not authored: the scenarios and the agent config
+already exist upstream, so this command dumps them into the packaged
+``contracts/`` tree where they become diffable, hashable artifacts that ship
+inside the wheel.
+
+Coval is always the source for the suite. A platform is optional and resolved
+through ``variants.platforms.FETCHERS``, so no vendor is special here and
+``--platform`` accepts any registered name.
+
+Where a dump lands matters. ``_source/`` holds provenance that is committed;
+``_private/`` holds the evaluator and is never committed. Promoting either into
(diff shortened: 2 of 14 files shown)


runner/tests/test_contracts.py
```diff
@@ -0,0 +1,80 @@
...
+    assert stack.tts.voice_id == "EXAVITQu4vr4xnSDxMaL"
+    assert (stack.media.codec, stack.media.sample_rate_hz) == ("PCMU", 8000)
+    assert stack.turn_taking.end_of_turn_target_ms == 800
+    # Both must be off or the benchmark measures the vendor's rubric and timer.
+    assert stack.platform_behaviour.native_auto_hangup is False
+    assert stack.platform_behaviour.vendor_post_call_analysis is False
+
+
+def test_contract_hash_is_pinned() -> None:
+    assert contract_sha256("dental").startswith(EXPECTED_CONTRACT_SHA), (
```
callumreid, line 47: **P2 — this pin fails for exactly the people who hold the evaluator, and the guard written for it is never called.**

`contract_sha256` hashes `PRIVATE_CONTRACT_FILES = ("_private/mock-tools.json",)` when the file is present (`contracts/__init__.py:76`, `:172-176`). I ran it in this checkout: `has_private_contract('dental') == False` and the digest is `146ac237d7ef007cb2d2aa32bdce408b…`, i.e. `EXPECTED_CONTRACT_SHA` was computed in the *fixtures-absent* state.

**Failure scenario:** a maintainer who actually has `contracts/dental/_private/mock-tools.json` — the working state for anyone who can run the suite — runs `uv run pytest`. The extra `digest.update` changes the hash, this assert fires, and the message says *"The dental contract or the pinned stack changed. If that was deliberate, update EXPECTED_CONTRACT_SHA in the same commit."* That is false, and following it commits a constant CI can never reproduce (`_private/` is gitignored, so CI is always in the absent state) — flipping the red from local to CI.

Existing validation does not catch it because CI runs from a fresh clone where the private file never exists; the 7 passing tests here prove only the absent branch.

`has_private_contract()` is defined and exported for precisely this and has **zero call sites** — and the module docstring at `contracts/__init__.py:73-76` already states the intended behavior: *"which is why the pinned test skips rather than fails when the directory is absent."* The code never skips. Smallest fix is to make the dead guard live:

```python
@pytest.mark.skipif(
    has_private_contract("dental"),
    reason="EXPECTED_CONTRACT_SHA is pinned to the fixtures-absent state",
)
def test_contract_hash_is_pinned() -> None:
```

(Or drop `PRIVATE_CONTRACT_FILES` from the digest if a state-independent hash is what you want — but that loses the "identifies which fixtures a run used" property the docstring argues for.)

runner/src/coval_bench/variants/pull.py
```diff
@@ -0,0 +1,233 @@
...
+    source_dir = out_root / "dental" / "_source"
+    private_dir = out_root / "dental" / "_private"
+    prompts: dict[str, str] = {}
+
+    with _coval(coval_api_key) as client:
+        click.echo(f"Coval agent {coval_agent_id}")
+        response = client.get(f"/agents/{coval_agent_id}")
+        response.raise_for_status()
+        agent = cast("dict[str, Any]", response.json()).get("agent", {})
+        _write(source_dir / "coval-agent.json", agent)
```
callumreid: **P2 — the Coval agent is dumped unredacted into a committed directory; the platform dump on the same command is redacted.**

Line 201 sends the platform config through `fetch_platform` → `redact()` before writing. This line writes the Coval agent verbatim. `_source/` is committed on purpose (README:79-86, and `.gitignore` excludes only `*/_private/`), so whatever comes back lands in git.

**Nothing leaked here** — I read the committed `_source/coval-agent.json`; it is a SIP/Vapi `MODEL_TYPE_VOICE` agent with `endpoint: ""` and no credential-shaped fields, matching the scan noted at README:106-108. The defect is that the command is documented as vendor-neutral and reusable for any agent id, and Coval's own agent schema carries key material for other model types. From `docs/api/openapi/agents-v1.yaml` on backend `origin/main`: `metadata.pipecat_api_key` ("Pipecat Cloud API key", required for `MODEL_TYPE_DAILY`), plus `custom_headers`, `trigger_call_headers`, `generate_token_headers`, `http_headers`, `voice_http_headers`.

**Failure scenario:** `coval-bench pull-contract --coval-agent-id <a Pipecat or websocket agent>` writes `metadata.pipecat_api_key` and any `Authorization` header value straight into `contracts/<suite>/_source/coval-agent.json`, which the next commit publishes to a public Apache-2.0 repo. Never committing secrets is a hard org rule, and the machinery to prevent it is already imported three lines away.

Fix — reuse the existing redactor:

```python
from coval_bench.variants.platforms import FETCHERS, fetch_platform, redact
...
agent_found: list[str] = []
_write(source_dir / "coval-agent.json", redact(agent, agent_found))
```

Two notes on the redactor while you are there: `SECRET_KEY` matches on key *name*, so a container key like `custom_headers` passes through and only its nested `Authorization`/`token` children get caught — worth matching header-container keys too. And `redact()` currently has no test at all, in a PR that otherwise pins a hash against accidental change; a couple of cases (nested dict, list, `credentialIds` deliberately untouched) would be cheap insurance on the one function whose bug is a published credential.

runner/src/coval_bench/contracts/README.md
```diff
@@ -0,0 +1,114 @@
...
+Promoting a `_source/` file into a contract file is a deliberate human step, so a
+re-pull can never silently rewrite a contract that a published run already used.
+
+### What is published, and what is not
+
+The agent definition is public; the evaluator is not.
+
+| File | Published | Why |
+|---|---|---|
+| `system-prompt.txt`, `first-message.txt`, `tool-definitions.json`, `mock-tools.json` | yes | This is the agent under test. A reader must be able to audit what the agent was told in order to trust a result |
```
callumreid: **P2 — the README tells a future contributor to put the evaluator fixtures at a path `.gitignore` does not cover, and marks them "Published: yes".**

`mock-tools.json` is claimed at three different locations in this PR, and only the third matches the code:

| Source | Claimed path |
|---|---|
| `stack.json:23` `keyterms_source` | `agent/dental/mock-tools.json` |
| this README (layout :24, table :94) | `dental/mock-tools.json`, **published** |
| the code (`contracts/__init__.py:76`, `:187`) | `dental/_private/mock-tools.json` |

I checked: `find runner -name 'mock-tools.json' -o -path '*agent/dental*'` returns nothing, so `keyterms_source` is a dangling pointer frozen into the hashed stack — correcting it later costs an `EXPECTED_CONTRACT_SHA` bump.

**The dangerous one is the table row.** `.gitignore` protects `runner/src/coval_bench/contracts/*/_private/` and nothing else. A contributor who follows the layout diagram and drops fixtures at `contracts/dental/mock-tools.json` gets them committed and published to a public repo — the exact outcome the `.gitignore` comment and `contracts/__init__.py:22-24` say must never happen ("Publishing the answers would let a platform optimise for the test rather than the task"). No check catches this: `git add` succeeds, CI has no test for it, and the contract hash is indifferent to which path the file sits at.

Same staleness one row up: `_source/test-cases.json | no, gitignored` — `pull.py:186` writes test cases to `_private/test-cases.json`, and `_source/` is *not* gitignored. And README:27 says `platforms/vapi-assistant.json` where `pull.py:201` writes `platforms/{platform}-agent.json`. This all reads like the final commit ("refactor(contracts): hide the evaluator") moved the evaluator under `_private/` without the README following.

callumreid (approved):
**APPROVE.** Three P2s inline, none blocking.

**Reviewed:** base `dac88f8481b3` → head `4414828885da3f06fbe0c74949237aa9c40a9892`, scoped three-dot. 16 files, +1020/-0, matching `changedFiles`. Base is `main`, not a stacked branch.

The design is careful and the security-sensitive parts are mostly right: `COVAL_API_KEY` and platform keys are read from the environment, never committed, and added to `config.py` as `SecretStr`; the `--platform` value is constrained by `click.Choice(sorted(FETCHERS))` so the only interpolation into a write path cannot traverse; every request goes over TLS to a hardcoded `api.coval.dev` / `api.vapi.ai` base with a 30s timeout and `raise_for_status()`; `_paginate` is bounded at `MAX_PAGES = 50` and raises rather than spinning. The command only reads remote APIs and writes JSON under `--out-root` — nothing it fetches is executed, installed, or evaluated, so the highest-risk shape for a `pull`-style command does not apply here.

**On "does the pin actually pin" — today it declares rather than enforces.** `load_stack()` / `stack_as_dict()` have no non-test callers, and `stack.json:5-7` asserts that "`plan` refuses to launch when a platform's live config diverges" — there is no `plan` command in this branch or on `main` (`__main__.py` registers `pull-contract` and `fetch-s2s` only). That reads as staged scaffolding rather than an overclaim and the enforcement is clearly a follow-up, so I am not blocking on it. Worth knowing that a reader of `stack.json` will currently believe divergence is checked when nothing checks it yet. What the pin *does* deliver is real: `_Annotated` rejects any non-`_` extra key, so a typo'd pin fails loudly instead of silently reverting to a vendor default, and that path is well tested.

**Validation actually run in a worktree at this head:** `uv run pytest tests/test_contracts.py -q` → 7 passed; `uv run ruff check --no-fix` on `contracts/`, `variants/`, `tests/test_contracts.py` → clean; `uv run ruff format --check` → 7 files already formatted; `uv run mypy src/coval_bench/contracts src/coval_bench/variants` → success, 6 files. `pull-contract` was deliberately **not** executed against any live API. The Coval agent schema claim in the second finding was confirmed against `docs/api/openapi/agents-v1.yaml` on backend `origin/main`.

**Quality gates at this exact SHA** (`check-runs` for `4414828885da`, not an aggregate): Switchboard Gate ✅, Benchmarks Runner CI Adapter / Runner CI ✅, CodeQL ✅ ("No new alerts in code changed by this pull request"), Plan checks ✅, Analyze Python ✅.

**No bot has actually reviewed this head.** CodeRabbit's green status text is "Review rate limited" (org quota — not a review, and not even among the check-runs for this SHA); Greptile posted "Your trial has ended"; the Codex review is against commit `094ae0fd62`, not `4414828885da`. This human pass is the only review this head has had.

**Staleness (a separate question from the diff, deliberately kept apart):** `main` is **48 commits** ahead of the merge base — the largest drift in today's queue. `gh` reports `MERGEABLE` and I confirmed no textual conflict; the only files touched by both sides are `.gitignore` and `runner/src/coval_bench/config.py`, in non-overlapping hunks. Nothing on `main` duplicates the contracts package. Two adjacent-but-not-conflicting commits worth a glance: `6c770ea` (#543, same author) gave dental appointment runs their own S2S family, so a dental test set is now referenced from the S2S fetch path *and* from this contract — worth checking whether `test-set 7rJxSKMp` here is the same set #543 points at, since two places now name it. `648d59a` (#534, COVAL-5297) removed *runner build SHA* provenance from results while this PR introduces `contract_sha256` provenance; different things, not a reversal, but the methodology doc #534 rewrote is where a contract hash would need describing. One rebase-mechanical note: `main` reorganized `runner/tests/` into `api/`, `providers/`, `runner/`, `unit/`; this PR adds `runner/tests/test_contracts.py` at top level. It is still collected (CI runs bare `uv run pytest -q` with no `testpaths`), so that is placement only, not a coverage gap.

**Residual risk: medium-low.** No migration, no schema change, no existing behavior touched — additive scaffolding behind a manually invoked CLI with no scheduled or automated caller, so the blast radius of a mistake is a developer's working tree rather than a published number. The two unquantified parts: the redaction path has no test and its failure mode is a committed credential, and the mock-tool router that justifies the whole package layout does not exist yet, so the `_private/` + `importlib.resources` arrangement is unproven at runtime (a gitignored `mock-tools.json` will not be in the `Dockerfile.api` build context when that router lands — worth designing for before it is written). Confidence in the three findings: high; each verified by reading the code and, where applicable, running it.

PR #138, opened 2026-06-17
Arena DB access layer and seed script

seribaymadina:
Layer 2 (+1) of the Voice Arena plan: the typed data-access layer for the arena schema, plus a local seed fixture. Stacks on feat/voice-arena.

- `arena/models.py` — Pydantic `Battle`/`Vote` (mirrors `db/models.py`); outcome/voter_type as `StrEnum` matching the DB checks.
- `arena/store.py` — async `ArenaStore` over the shared pool: `insert_battle`, `upsert_vote` (ON CONFLICT → updates the existing vote; the trigger bumps `updated_at`), `get_battle`, `list_battles`, `list_votes`.
- `arena/seed.py` — `python -m coval_bench.arena.seed`: 4 models, 6 battles (all pairs = connected graph), 7 deterministic votes each. Skips if already seeded; `--reset` to reseed. Local dev only.
- tests — embedded Postgres; covers vote dedup-as-update, the self-battle check, and domain/battle scoping.

Validated: ruff, ruff format, mypy --strict, 5/5 tests; seed verified against the local DB (6 battles, 42 votes).

No API/web/rating/audio changes.

## Release Notes

* **New Features**
  * Introduced arena system for model-to-model comparisons with voting-based evaluation.
  * Implemented persistent storage for battles, votes, and voting outcome tracking.

* **Tests**
  * Added comprehensive test coverage for arena battles and voting persistence layer.

<h3>Greptile Summary</h3>

Adds the typed data-access layer (`ArenaStore`) and Pydantic models (`Battle`, `Vote`, `VoteOutcome`, `VoterType`) for the voice arena schema, along with a local seed script and embedded-Postgres tests. All SQL is parameterised, transaction management is correct (explicit `conn.commit()` for writes, no-op for reads), and the seed script includes a conservative `_assert_local` guard that blocks execution against any non-loopback host.

- `arena_store.py` exposes `insert_battle`, `upsert_vote` (ON CONFLICT → update), `get_battle`, `list_battles`, and `list_votes`; the `list_votes` unbounded-by-default behaviour is intentional and documented (rating refit needs all votes).
- `seed_arena.py` seeds 4 models × 6 pairings × 7 deterministic votes and is idempotent (skip-if-seeded / `--reset`); the `_clear` helper deletes in FK-safe order (`leaderboard_snapshots → votes → battles`).
- Five tests cover the full read/write round-trip, vote dedup-as-update, self-battle DB check constraint, domain filtering, and cross-battle vote scoping.

[...]

PR #547, opened 2026-08-25
Bound benchmark dashboard API payloads

kdmelon:
## Summary

- add backward-compatible `include_series=false` support to aggregate requests and skip series SQL entirely on that path
- add a compact `/v1/results/timeline` response for dashboard charts
- keep 24h and 7d timelines at full bucket resolution
- bound 30d timelines in PostgreSQL to at most 240 representative points per provider/model/metric while preserving first, last, and plotted extrema
- apply the same 30d bound to the legacy aggregate series path
- keep internal-access visibility isolated in cache keys and add focused regression coverage

## Validation

- Ruff format and lint: PASS
- strict mypy on changed API source: PASS
- Python compile check: PASS
- `git diff --check`: PASS
- exact production SQL executed against PostgreSQL 16: two 241-point groups each returned 238 deterministic points, retaining endpoints and injected interior extrema; 7d returned every qualifying point with the compact projection
- independent Tier 3 review: PASS

Focused endpoint pytest was attempted, but the local host lacks the `pg_config` executable required by the pytest-postgresql fixture. The failure occurred during fixture setup rather than in a test assertion; the exact query received the PostgreSQL 16 smoke coverage described above.

## Rollout

Deploy this API change before the dependent benchmarks-web PR so `/v1/results/timeline` exists before the web client requests it.

Related: [infrastructure PR #138](https://github.com/coval-ai/benchmark-infra/pull/138) and [web PR #49](https://github.com/coval-ai/benchmarks-web/pull/49).

### DIFF
--- a/runner/src/coval_bench/api/routers/aggregates.py
+++ b/runner/src/coval_bench/api/routers/aggregates.py
@@ -59,6 +59,8 @@ from coval_bench.api.schemas import (
     DatasetAggregates,
     ModelStatEntry,
     SeriesPoint,
+    TimelinePoint,
+    TimelineResponse,
 )
 from coval_bench.config import DATASET_ALL
 from coval_bench.registries import is_metric_excluded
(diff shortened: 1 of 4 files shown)


coval-cale (approved):
Hell ya dude!

PR #548, opened 2026-08-26
[BENCH-734] Add Baseten qwen3-asr-1.7b STT and warm Baseten endpoints…

coval-cale:
… before runs

qwen3-asr-1.7b joins as an early-access dedicated entry speaking the input_audio_buffer wire protocol (base64 JSON appends, commit to finalize, is_end_of_audio_flush to end); the whisper deployments keep the binary 512-sample protocol, verified against all three endpoints. Endpoint URLs resolve per model, so whisper-large-v3's cutover to the v2 deployment is a secret rotation with no registry change.

Baseten is retiring its own warmup traffic in favor of demand autoscaling, making our warmup the scale-up trigger: both Baseten providers now warm every configured endpoint concurrently with a 6-minute budget (a scale-from-zero boot measured ~166 s), run_probe warms the same way run_benchmarks does, warmup timeouts are logged rather than swallowed, and the tone clip's expected close-without-final is not reported as an error.

### DIFF
--- a/docs/methodology.md
+++ b/docs/methodology.md
@@ -161,10 +161,13 @@ non-persisting probe (`coval-bench probe`) warms the same way, so its numbers
 are comparable to a scheduled run's.
 
 Dedicated endpoints warm differently: rather than a connection, what needs
-warming is the replica behind it. The Baseten STT provider streams a one-second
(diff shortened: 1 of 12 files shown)


seribaymadina (approved):
nice!

PR #550, opened 2026-08-26
[BENCH-737] Add Zoom Scribe live STT as an early-access entry

coval-cale:
Scribe live is a WebSocket session (subprotocol live-asr) authenticated with a per-connection HS256 JWT signed from the Zoom Build key/secret pair, so the provider takes two credentials and the orchestrator passes the secret alongside the conventional api_key.

The wire protocol emits only turn-final transcription.completed events: the documented enable_partial_results flag silently disables transcription (verified live), and session.close drops a pending turn instead of flushing it, so the client pads up to 2 s of trailing silence until the endpointer fires, as revai and together already do. TTFT and TTFS are excluded accordingly: the first token is the turn-final, and final timing bundles endpoint detection. WER, AudioToFinal, and RTF publish; a 5-item live probe measured 2.1% mean WER.

Non-fatal error events are logged and skipped since Zoom uses them as warnings (e.g. config hints); only errors flagged fatal end a run.

### DIFF
--- a/runner/src/coval_bench/providers/stt/zoom.py
+++ b/runner/src/coval_bench/providers/stt/zoom.py
@@ -1,0 +1,223 @@
+# Copyright 2026 The Coval Benchmarks Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Zoom Scribe live STT provider (WebSocket)."""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import time
+from typing import Any
+
+import jwt
+import structlog
+import websockets.asyncio.client as ws_client
+from pydantic import SecretStr
+from websockets.typing import Subprotocol
+
+from coval_bench.providers.base import STTProvider, TranscriptionResult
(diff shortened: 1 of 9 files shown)


runner/src/coval_bench/providers/stt/zoom.py
```diff
@@ -0,0 +1,223 @@
...
+    def _auth_token(self) -> str:
+        # iat back-dated 30 s for clock skew
+        iat = int(time.time()) - 30
+        return jwt.encode(
+            {"iss": self._api_key.get_secret_value(), "iat": iat, "exp": iat + _TOKEN_TTL_S},
+            self._api_secret.get_secret_value(),
+            algorithm="HS256",
+        )
+
+    async def measure_ttft(
```
seribaymadina, line 70: correct me if Im wrong but since it cant emit partials and ttft is excluded anyways & isnt needed, no?

seribaymadina (approved):
left one non blocker

PR #553, opened 2026-08-26
Add the result variant dimensions and the mock tool-call log

seribaymadina:
Schema only, nothing runs yet.

Orchestration platforms measure under the existing `S2S` benchmark, not one of their own — same metrics off the same recording, so the platform is a dimension of a row. `variant_id` is that dimension, NOT NULL defaulting to `pinned`, which is also the backfill since every existing row is Coval-pinned. `transport` and `test_case_id` are nullable and can't be reconstructed once a run stops being queryable, so this wants to land before the first row worth keeping. No `iteration` column — it falls out of `row_number() over (partition by run_id, variant_id, test_case_id order by created_at, id)`.

`mock_tool_calls` is the tool-call ground truth, indexed on `(simulation_id, created_at, id)` for the ingest join and `(caller_number, created_at)` for when a platform drops the `X-Coval-Simulation-Id` SIP header. The trailing `id` matters: `now()` is transaction start, so one transaction's calls share a `created_at`. Separate table because a tool call is an event with arguments and a response, not a numeric metric, and the rollups filter on `metric_value` being non-null.


coval-cale (approved):
Sick!

PR #568, opened 2026-08-31
Add LLM chat benchmark via Coval MODEL_TYPE_CHAT simulator

bnhopkins:
## Summary
- Adds a new `LLM` benchmark that fetches instruction-following scores from Coval's chat simulator (`MODEL_TYPE_CHAT`), mirroring the S2S fetch-and-ingest pattern
- Measures instruction adherence identically to S2S (YES→100, NO→0, UNKNOWN→skip, aggregated as pass rate)
- Registers 4 initial models: `openai/gpt-4o`, `openai/gpt-4o-mini`, `anthropic/claude-sonnet-4-6`, `google/gemini-2.5-flash`
- Wires up `coval-bench fetch-llm` CLI command with backfill support (`--coval-run-id`)

## Test plan
- [ ] Verify `python -m coval_bench fetch-llm --help` shows the command
- [ ] Set `COVAL_LLM_*` env vars and run `fetch-llm` against a test Coval agent with `MODEL_TYPE_CHAT`
- [ ] Confirm instruction-following rows land in the DB with `benchmark=LLM`
- [ ] Test targeted backfill with `--coval-run-id`

### DIFF
--- a/runner/src/coval_bench/__main__.py
+++ b/runner/src/coval_bench/__main__.py
@@ -23,6 +23,7 @@ from coval_bench.db.cli import db_check, db_migrate
 from coval_bench.migrations.backfill_normalized_storage import backfill_normalized_storage_cli
 from coval_bench.migrations.backfill_wer_breakdown import backfill_wer_breakdown_cli
 from coval_bench.migrations.import_legacy import import_legacy_cli
+from coval_bench.llm.fetch_chat import fetch_llm
 from coval_bench.s2s.fetch_v2v import fetch_s2s
 from coval_bench.variants.pull import pull_contract
 
@@ -108,8 +109,9 @@ migrate.add_command(backfill_wer_breakdown_cli, name="backfill-wer-breakdown")
 migrate.add_command(backfill_normalized_storage_cli, name="backfill-normalized-storage")
 migrate.add_command(import_legacy_cli, name="import-legacy")
 
-# S2S is fetch-only, so a standalone command rather than a `run --kind` value.
+# S2S and LLM are fetch-only, so standalone commands rather than `run --kind` values.
 cli.add_command(fetch_s2s, name="fetch-s2s")
+cli.add_command(fetch_llm, name="fetch-llm")
 cli.add_command(pull_contract, name="pull-contract")
 
 
--- a/runner/src/coval_bench/llm/fetch_chat.py
+++ b/runner/src/coval_bench/llm/fetch_chat.py
@@ -1,0 +1,604 @@
+# Copyright 2026 The Coval Benchmarks Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Fetch LLM chat instruction-following scores from the Coval API.
+
+Ingests each provider's recent completed runs (MODEL_TYPE_CHAT agents) not yet
+in the DB, slotted by run create_time. Agent ids, the metric id, and the API
+key are read from the environment, never committed.
+
+The structure mirrors ``s2s.fetch_v2v`` but is simpler: no latency metric, no
+caller-condition splits, no sample publishing. The single metric is instruction
+following, scored identically to the S2S benchmark (YES/NO/UNKNOWN per
+conversation, aggregated as a pass rate).
+"""
+
+from __future__ import annotations
+
+import asyncio
+import hashlib
+import importlib.resources
+from collections.abc import Callable
+from dataclasses import dataclass
+from datetime import UTC, datetime
+from typing import Any, cast
+
+import click
+import httpx
+import structlog
+
+from coval_bench.config import Settings, get_settings
+from coval_bench.db.conn import lifespan_pool
+from coval_bench.db.models import Result, ResultStatus, RunStatus
+from coval_bench.db.writer import RunWriter
+from coval_bench.registries import METRIC_SPECS, Metric
+from coval_bench.registries.benchmarks import Benchmark
+
+logger = structlog.get_logger("coval_bench.llm.fetch_chat")
+
+DATASET_ID = "llm-v1"
+WINDOW_FLOOR_SECONDS = 86_400
+WINDOW_PAGE_SIZE = 10
+
+
+@dataclass(frozen=True)
+class AgentSpec:
+    """One LLM chat provider: the Settings attr holding its Coval agent id + display strings."""
+
+    agent_id_attr: str
+    provider: str
+    model: str
+
+
+@dataclass(frozen=True)
+class CovalRun:
+    """One completed Coval run from the list endpoint's summary view."""
+
+    run_id: str
+    create_time: datetime | None
+
+
+AGENTS: tuple[AgentSpec, ...] = (
+    AgentSpec(
+        agent_id_attr="coval_llm_openai_agent_id",
+        provider="openai",
+        model="gpt-4o",
+    ),
+    AgentSpec(
+        agent_id_attr="coval_llm_openai_mini_agent_id",
+        provider="openai",
+        model="gpt-4o-mini",
+    ),
+    AgentSpec(
+        agent_id_attr="coval_llm_anthropic_agent_id",
+        provider="anthropic",
+        model="claude-sonnet-4-6",
+    ),
+    AgentSpec(
+        agent_id_attr="coval_llm_google_agent_id",
+        provider="google",
+        model="gemini-2.5-flash",
+    ),
+)
+
+
+def _client(settings: Settings) -> httpx.AsyncClient:
+    """Build the Coval API client."""
+    key = settings.coval_api_key
+    if key is None:
+        raise RuntimeError("coval_api_key is not set (Secret Manager in prod, .env locally)")
+    return httpx.AsyncClient(
+        base_url=settings.coval_api_base,
+        headers={"X-API-Key": key.get_secret_value()},
+        timeout=30.0,
+    )
+
+
+def _dataset_sha256() -> str:
+    """SHA-256 of the packaged LLM manifest (or 'unknown' if not yet created)."""
+    try:
+        ref = importlib.resources.files("coval_bench.datasets.manifests").joinpath(
+            f"{DATASET_ID}.json"
+        )
+        return hashlib.sha256(ref.read_bytes()).hexdigest()
+    except Exception:
+        return "unknown"
+
+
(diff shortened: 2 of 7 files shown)


runner/src/coval_bench/llm/fetch_chat.py
```diff
@@ -0,0 +1,604 @@
...
+        log_run_failed(str(exc), exc)
+        raise
+
+    failed = [p for p, s in statuses.items() if s is RunStatus.FAILED]
+    if not statuses or all(s is RunStatus.FAILED for s in statuses.values()):
+        if statuses:
+            log_run_failed(f"llm fetch failed for all providers: {', '.join(failed)}")
+        else:
+            log_run_failed("llm fetch ran no providers (none configured)")
+        raise click.ClickException("llm fetch failed for all providers")
```
kobicovaldev, line 598: Curious about the page size logic here:

```python
page_size=page_size if only_run_ids else WINDOW_PAGE_SIZE,
```

When `--coval-run-id` is passed, `only_run_ids` is truthy, so this resolves to `page_size` (the CLI default of 100). But what if someone passes `--page-size 50`? That value gets shadowed by the CLI default of 100 before it reaches this line.

More importantly: is the intent here to use the user-specified page size during backfill, or to force `WINDOW_PAGE_SIZE` (10) for backfill searches? If it's the former, the conditional seems backwards — it'd be `page_size=page_size` regardless. If the latter, the CLI `--page-size` flag is misleading since it only applies to the non-backfill path.

Either way, the current behavior means backfill always uses page size 10, which is 10× slower than the documented default. Worth a quick sanity check?

runner/src/coval_bench/llm/fetch_chat.py
```diff
@@ -0,0 +1,604 @@
...
+
+
+def _dataset_sha256() -> str:
+    """SHA-256 of the packaged LLM manifest (or 'unknown' if not yet created)."""
+    try:
+        ref = importlib.resources.files("coval_bench.datasets.manifests").joinpath(
+            f"{DATASET_ID}.json"
+        )
+        return hashlib.sha256(ref.read_bytes()).hexdigest()
+    except Exception:
```
kobicovaldev, line 104: [Suggestion] This bare `except Exception` swallows any failure to read the manifest (a typo in the path, a missing file, a permissions error) and persists the string `"unknown"` as the dataset sha256, which future queries will treat as a real value. Would it be clearer to follow the `maybe_` convention here, e.g. `maybe_dataset_sha256() -> str | None` that returns `None` on failure, so callers can decide whether to log or fall back? At minimum, a `logger.debug` or `logger.warning` in the except block would make the sentinel traceable.

runner/src/coval_bench/llm/fetch_chat.py
```diff
@@ -0,0 +1,604 @@
...
+
+def _client(settings: Settings) -> httpx.AsyncClient:
+    """Build the Coval API client."""
+    key = settings.coval_api_key
+    if key is None:
+        raise RuntimeError("coval_api_key is not set (Secret Manager in prod, .env locally)")
+    return httpx.AsyncClient(
+        base_url=settings.coval_api_base,
+        headers={"X-API-Key": key.get_secret_value()},
+        timeout=30.0,
```
kobicovaldev, line 93: [Suggestion] This is a timeout in seconds, but nothing in the signature says so. Would a named constant such as `client_timeout_seconds = 30.0` (module level, next to `WINDOW_PAGE_SIZE`) make the unit obvious to the next reader?

runner/src/coval_bench/llm/fetch_chat.py
```diff
@@ -0,0 +1,604 @@
...
+
+    settings = get_settings()
+    configure_logging(level=settings.log_level)
+    try:
+        only_run_ids = frozenset(coval_run_ids) or None
+        statuses = asyncio.run(
+            fetch_and_write_llm(
+                settings,
+                only_run_ids=only_run_ids,
+                window_seconds=window_hours * 3600 if only_run_ids else None,
```
kobicovaldev, line 584: [Suggestion] `window_hours * 3600` hides a unit conversion in the call site. Would a named constant like `SECONDS_PER_HOUR = 3600` (or `HOURS_TO_SECONDS`) keep the math self-documenting, in line with the `_seconds` suffix convention used elsewhere in this file?

runner/src/coval_bench/llm/fetch_chat.py
```diff
@@ -0,0 +1,604 @@
...
+            "order_by": "-create_time",
+            "page_size": page_size,
+        }
+        if page_token:
+            params["page_token"] = page_token
+        resp = await client.get("/runs", params=params)
+        resp.raise_for_status()
+        payload = cast("dict[str, Any]", resp.json())
+        raw = cast("list[dict[str, Any]]", payload.get("runs", []))
+        for r in raw:
```
kobicovaldev, line 194: [Nit] Text is cheap: would `for run in raw:` be clearer than `for r in raw:`? The loop body builds a `run` right away, so the short name saves little.

runner/src/coval_bench/llm/fetch_chat.py
```diff
@@ -0,0 +1,604 @@
...
+    spec: AgentSpec,
+) -> list[Result]:
+    """Map per-conversation instruction-following values to Result rows."""
+    rows: list[Result] = []
+    for i, v in enumerate(values):
+        mapped = _VALUE_MAPPER(v.get("value"))
+        if mapped is None:
+            continue
+        metric_value, status = mapped
+        sim_id = v.get("simulation_output_id")
```
kobicovaldev, line 235: [Nit] Would `simulation_output_id` (the name of the field it reads, and the one used in the `audio_filename` path a few lines below) be clearer than `sim_id`?

runner/src/coval_bench/llm/fetch_chat.py
```diff
@@ -0,0 +1,604 @@
...
+logger = structlog.get_logger("coval_bench.llm.fetch_chat")
+
+DATASET_ID = "llm-v1"
+WINDOW_FLOOR_SECONDS = 86_400
+WINDOW_PAGE_SIZE = 10
+
+
+@dataclass(frozen=True)
+class AgentSpec:
+    """One LLM chat provider: the Settings attr holding its Coval agent id + display strings."""
```
kobicovaldev, line 46: [Nit] This docstring is a noun phrase rather than a complete sentence, and `attr` is an abbreviation (`attribute`). A few of the new docstrings share this style (the `CovalRun` docstring, `"""SHA-256 of the packaged LLM manifest..."""`, `"""YES -> 100.0, NO -> 0.0, UNKNOWN -> no row."""`, and the `recent_completed_runs` docstring). Would phrasing them as full sentences, e.g. `"""One LLM chat provider: the Settings attribute holding its Coval agent id, plus display strings."""`, keep documentation generation and grepping consistent across the file?

kobicovaldev (approved):

runner/src/coval_bench/llm/fetch_chat.py
```diff
@@ -0,0 +1,604 @@
...
+            page_size=page_size,
+            requested_run_ids=only_run_ids,
+        )
+        data_seen = False
+        newest_data_at: datetime | None = None
+        for coval_run in runs:
+            if only_run_ids is not None and coval_run.run_id not in only_run_ids:
+                continue
+
+            already = await writer.coval_metric_ingested(
```
callumreid, line 408: **P1 - LLM ingestion is not idempotent.** This calls `RunWriter.coval_metric_ingested`, but that helper is hard-coded to `r.benchmark = 'S2S'`. LLM rows are written with `Benchmark.LLM`, so already-ingested LLM runs will never be detected and each scheduled fetch can duplicate the same Coval run into fresh result rows, corrupting aggregates. Please make the idempotency check benchmark-aware or add an LLM-specific check before publishing this benchmark.

runner/src/coval_bench/__main__.py
```diff
@@ -108,8 +109,9 @@ def migrate() -> None:
 migrate.add_command(backfill_normalized_storage_cli, name="backfill-normalized-storage")
 migrate.add_command(import_legacy_cli, name="import-legacy")
 
-# S2S is fetch-only, so a standalone command rather than a `run --kind` value.
+# S2S and LLM are fetch-only, so standalone commands rather than `run --kind` values.
 cli.add_command(fetch_s2s, name="fetch-s2s")
+cli.add_command(fetch_llm, name="fetch-llm")
```
callumreid, line 114: **P1 - the paired infra invokes a command this runner does not register.** The new command is `fetch-llm`, while coval-ai/benchmark-infra#183 configures the production job as `sync-llm`. At this exact head, `python -m coval_bench fetch-llm --help` works and `python -m coval_bench sync-llm --help` fails, so the deployed Cloud Run job would fail at startup. Please align the command name with the infra before merging the pair.

runner/src/coval_bench/__main__.py
```diff
@@ -23,6 +23,7 @@
 from coval_bench.migrations.backfill_normalized_storage import backfill_normalized_storage_cli
 from coval_bench.migrations.backfill_wer_breakdown import backfill_wer_breakdown_cli
 from coval_bench.migrations.import_legacy import import_legacy_cli
+from coval_bench.llm.fetch_chat import fetch_llm
```
callumreid, line 26: **P2 - focused runner lint fails at the exact head.** Import order in this file is not sorted under the repo ruff settings; `uv run ruff check ...` fails here, matching the failing Runner CI status. Please sort the imports so the exact-head quality gate can pass.

callumreid (changes requested):
REQUEST_CHANGES. Reviewed exact head 0b1e3fe52b6504c16290f2921d6920d2e21460b8 against fresh origin/main and compared it with coval-ai/benchmark-infra#183 at 83382d2a99203ad58dd109e69c6a27ea12e06d68. The new LLM benchmark scaffolding is directionally coherent, `fetch-llm --help` is exposed locally, `Metric.INSTRUCTION_FOLLOWING` is registered for `Benchmark.LLM`, and the registry rows match this PR's provider/model strings. Blocking on LLM ingestion idempotency, the paired infra command/env mismatch, and a focused ruff failure. I did not run live Coval fetches or DB writes.

coval-cale:
Superseded by the unified LLM sync work (BENCH-811, 814, 827, 861).

PR #570, opened 2026-09-01
BENCH-768: Bound normalized readiness checks

kdmelon:
## Summary
- replace the unbounded seven-day bidirectional EXCEPT ALL query with exact grouped-count comparisons over bounded time/cohort chunks
- accumulate signed multiplicities across the full snapshot so duplicates and cross-chunk matches retain exact whole-window semantics
- use benchmark/dataset-constrained reads, conservative index-seeking coverage proof, per-statement timeouts, and a total deadline
- preserve TTS dataset coercion, TTFA public metric names, WER components, rollup parity, and machine-readable 0/1/2 outcomes

## Validation
- 21 focused readiness tests pass
- Ruff formatting and lint pass for all changed files
- strict mypy passes with explicit package bases
- independent high-risk review approved with no remaining findings
- added embedded-Postgres coverage for duplicate WER rows across chunks, TTS/TTFA mapping, WER components, and rollups; local execution is blocked only because this host has no Postgres binaries, so CI must exercise it

## Rollout safety
- no schema migration and no read-path flag change
- normalized dashboard reads remain disabled
- after deployment, first run EXPLAIN without ANALYZE on representative chunks, then run the bounded checker once
- cut over only if the complete checker exits 0

Related: BENCH-768

### DIFF
--- a/runner/scripts/check_normalized_readiness.py
+++ b/runner/scripts/check_normalized_readiness.py
@@ -3,10 +3,10 @@
 
 """Read-only readiness gate for normalized dashboard reads.
 
-The checker takes one repeatable-read, read-only snapshot, fixes an ``as_of``
+The checker takes one repeatable-read, read-only snapshot, fixes an as_of
 cutoff, and reports whether eligible legacy data spans at least seven days and
 the latest seven-day window has exact raw public-row and materialized-rollup
-parity. It never writes.
+parity. Every parity query is time-bounded and the checker never writes.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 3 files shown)


runner/scripts/check_normalized_readiness.py
```diff
@@ -129,68 +320,304 @@ def _coverage(buckets: Iterable[datetime]) -> tuple[float, datetime | None, date
...
+    counter: Counter[tuple[Any, ...]],
+    rows: Iterable[Sequence[Any]],
+    *,
+    sign: int,
+) -> None:
+    for row in rows:
+        if not row:
+            raise ReadinessOperationalError("parity", "count query returned an empty row")
+        *public_row, multiplicity = row
+        counter[tuple(public_row)] += sign * int(multiplicity)
```
coval-cale: [P1] Remove entries when their updated count reaches zero. `Counter` keeps zero-valued keys, so even a perfect legacy/normalized match retains every distinct public row for the full seven-day window. In a small reproduction, 100,000 matching rows left 100,000 keys and used about 18 MiB; millions of mostly unique metric values could consume hundreds of MiB and make this readiness check fail from memory pressure. Deleting `counter[key]` when it becomes zero preserves exact cross-chunk counting while releasing rows that have fully cancelled.

coval-cale (changes requested):
The chunked comparison and count-based matching look sound, and the focused tests, real-Postgres test, Ruff, mypy, and CI all pass. I found one scalability issue that should be fixed before merge: matching rows remain as zero-count keys in memory, so the checker can still retain essentially every distinct result from the seven-day window.

coval-cale (approved):

runner/scripts/check_normalized_readiness.py
kdmelon: Fixed in bb9b558c. Fully canceled keys are now deleted immediately, preserving exact Counter semantics while releasing matched rows. The cross-chunk cancellation regression test now also asserts that the Counter is empty; 21 focused tests, Ruff, and strict mypy pass.

PR #574, opened 2026-09-01
Manage the Vapi assistant's tools from the repo

seribaymadina:
The Vapi dental assistant had no tools. This puts them there from the repo.

- `platform-assets plan|apply --agent vapi-dental --mock-base-url …` — `plan` diffs, `apply` confirms then PATCHes
- Five tools rendered from `tool-definitions.json`: `server.url` → `/mock/vapi`, headers carry `X-Mock-Tools-Key` and `X-Coval-Simulation-Id: {{coval-simulation-id}}`, no `messages[]` so no filler speech
- Only `model.tools` is managed. PATCH sends the whole live `model` with tools replaced, so prompt and temperature are untouched
- One `Platform` row per vendor (renderer + client behind a two-method protocol); Telnyx is one more row
- `SecretRef` lands in a committed module; redactor now catches `X-Mock-Tools-Key`

Needs `VAPI_API_KEY`, `VAPI_DENTAL_ASSISTANT_ID`, `MOCK_TOOLS_SECRET` in env.


coval-cale (approved):
EPIC!!!!

PR #578, opened 2026-09-02
[BENCH-783] Add normalized S2S storage backfill

kdmelon:
## Summary
- add a standalone database-only normalized S2S storage backfill command
- reconcile deterministic backfill rows and random-ID live dual writes without fabricating artifact lineage
- verify fresh payloads, complete observation population, public primary-value parity, and exact rollups before reporting backfill_complete
- add structured phase/checkpoint progress, a disposable local benchmark harness, and the operator runbook

## Safety
- apply requires an explicit frozen maximum result ID
- dry run is read-only and can auto-freeze the S2S maximum
- writes commit one complete run-ID page at a time under the shared normalized-storage lock
- protected STT/TTS, live S2S dual-write, and readiness implementations are unchanged
- no production apply or dashboard-read rollout was performed

## Validation
- 18 focused S2S backfill tests passed against disposable PostgreSQL
- 86 existing normalized dual-write and S2S fetch tests passed
- 23 relevant sibling backfill tests passed
- Ruff format/check and mypy passed
- adversarial smoke confirmed an extra failed normalized observation blocks completion
- disposable benchmark smoke completed for multiple batch sizes and removed its database
- independent Tier 3 Sol review passed after one repair pass

Linear: https://linear.app/coval/issue/BENCH-783/implement-standalone-normalized-s2s-storage-backfill

### DIFF
--- a/runner/src/coval_bench/migrations/backfill_normalized_s2s_storage.py
+++ b/runner/src/coval_bench/migrations/backfill_normalized_s2s_storage.py
@@ -39,6 +39,7 @@ _FALLBACK_EVALUATION_ERROR = "legacy metric produced no value"
 _MISMATCH_DETAIL_LIMIT = 100
 _PROGRESS_INTERVAL_SECONDS = 30.0
 _PROGRESS_MAX_UNITS = 10
+_PUBLIC_PARITY_STATEMENT_TIMEOUT = "30s"
 _METRICS = {
     "V2V": "milliseconds",
     "InstructionFollowing": "percent",
@@ -747,7 +748,7 @@ FROM benchmarks_v2.metric_values v
 JOIN benchmarks_v2.metric_evaluations e ON e.id=v.metric_evaluation_id
 JOIN benchmarks_v2.benchmark_observations o ON o.id=e.observation_id
 JOIN benchmarks_v2.runs r ON r.id=o.run_id
-WHERE o.status='succeeded' AND e.status='succeeded'
+WHERE o.benchmark='S2S' AND o.status='succeeded' AND e.status='succeeded'
   AND r.status IN ('succeeded','partial') AND r.scheduled_at=%(bucket)s
 GROUP BY GROUPING SETS (
   (o.provider,o.model,o.benchmark,o.dataset_id,e.metric_type,e.metric_version,
@@ -760,7 +761,8 @@ _STORED_ROLLUP_SQL = """
 SELECT provider,model,benchmark,dataset_id,metric_type,metric_version,
        evaluation_variant,value_key,unit,bucket_at,min_value,p25,p50,p75,
        max_value,value_sum,sample_count
-FROM benchmarks_v2.metric_values_by_bucket WHERE bucket_at=%(bucket)s
+FROM benchmarks_v2.metric_values_by_bucket
+WHERE bucket_at=%(bucket)s AND benchmark='S2S'
 """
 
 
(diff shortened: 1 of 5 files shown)


ben-coval (approved):

runner/src/coval_bench/migrations/backfill_normalized_s2s_storage.py
```diff
@@ -0,0 +1,1153 @@
...
+ORDER BY 1
+"""
+
+
+def _verify_public_parity(
+    conn: psycopg.Connection, low: int, high: int, report: dict[str, Any]
+) -> None:
+    with conn.transaction(), conn.cursor() as cur:
+        cur.execute("SET TRANSACTION READ ONLY")
+        cur.execute(_PUBLIC_PARITY_SQL, {"low": low, "high": high})
```
coval-cale: [P1] Page the parity verification

`fetchmany(100)` only pages rows after PostgreSQL has executed the full query. `_PUBLIC_PARITY_SQL` still builds the entire frozen cohort, groups both complete legacy and normalized populations, performs a full outer join, and sorts mismatches in one statement. On a production-sized history this can consume substantial CPU, memory, temporary disk, and I/O despite the advertised 100-run page bound. Please verify parity per run-ID page and add a defensive statement timeout.

runner/src/coval_bench/migrations/backfill_normalized_s2s_storage.py
```diff
@@ -0,0 +1,1153 @@
...
+"""
+
+
+def _refresh(cur: psycopg.Cursor[Any], bucket: datetime) -> None:
+    params = {"bucket": bucket}
+    cur.execute(
+        "SELECT pg_advisory_xact_lock(hashtextextended('metric_values_by_bucket',extract(epoch FROM %(bucket)s::timestamptz)::bigint))",
+        params,
+    )
+    cur.execute(
```
coval-cale, line 775: [P1] Do not rebuild every benchmark rollup

This S2S-only migration deletes every normalized rollup for the bucket, including STT, TTS, and LLM rows, and then recomputes percentiles across every benchmark. It does this once per affected bucket per source page, creating unnecessary deletes, WAL, scans, and read churn. The advisory lock serializes competing refreshes but does not reduce that load. Please restrict both the DELETE and aggregate SELECT to `benchmark = 'S2S'`, or use a purpose-specific S2S refresh.

coval-cale (approved):
Approved with the database-safety follow-ups noted in the inline comments.

runner/src/coval_bench/migrations/backfill_normalized_s2s_storage.py
kdmelon: Fixed in fff17e06. Public parity now keyset-pages complete terminal S2S run IDs using the configured batch size, executes one bounded parity query per page, and installs a transaction-local 30s statement timeout before each query. Exact identity/value/multiplicity comparison is unchanged. Added real-Postgres coverage with batch_size=1 plus a timeout-order regression.

runner/src/coval_bench/migrations/backfill_normalized_s2s_storage.py
kdmelon: Fixed in fff17e06. The rollup aggregate, DELETE, and stored verification are now all scoped to benchmark='S2S' while retaining the existing per-bucket advisory lock. Added real-Postgres coverage proving same-bucket STT/TTS rollups remain byte-for-byte unchanged and are outside S2S verification, while exact S2S value_sum tampering is still detected.

PR #581, opened 2026-09-02
[BENCH-794] Refresh stats matviews once per scheduled window

coval-cale:
Every run rebuilt all three stats views, so nine runs per tick queued on the view locks and rode the queue into the Cloud Run task timeout. Now only the slot's last finisher refreshes, with an advisory try-lock covering two runs finishing together. The s2s fetch keeps its unconditional refresh.


callumreid (approved):
refreshing!

PR #585, opened 2026-09-02
[BENCH-785] Make the Coval fetch benchmark-aware

coval-cale:
Basically takes the current coval fetch job and generalizes it to the LLM benchmark as well.


seribaymadina (approved):
LGTM!

PR #591, opened 2026-09-03
[BENCH-797] Prevent STT/TTS dual-write gaps

kdmelon:
## Summary
- remove the application-level persistence semaphore; production evidence tied the observed pool timeouts to one concurrent DDL-lock incident, and the driver pool already queues acquisition
- retry only evidence-supported, replay-safe failures: legacy persistence retries `PoolTimeout`; normalized persistence retries `PoolTimeout` and `psycopg.OperationalError`
- keep normalized artifact upload outside retries, preserve a stable `captured_at`, and normalize blank provider errors
- add public pool-state diagnostics to retry/exhaustion logs so future incidents distinguish queued demand, blocked/unavailable connections, and connection failures
- leave S2S enablement and read-cutover behavior unchanged

## Incident evidence
- all 287 production `PoolTimeout` events found in the preceding 30 days occurred in one roughly 157-second window across four concurrent benchmark executions
- the same window contained the `20260901_0025_add_llm_benchmark.py` DDL deadlock; migration-locking remediation landed in #580 and #582
- separate production failures support the retained changes: 14 blank/whitespace provider-error validation failures and 5 normalized connection-loss `OperationalError` failures

## Testing
- `182 passed` — focused normalized dual-write, orchestrator, and S2S unit tests
- Ruff check and format check passed on all nine changed files
- `mypy --strict src tests` passed across 322 source files
- `git diff --check` passed
- 48 PostgreSQL writer tests collect, including composed cases for failure before observation acquisition and ambiguous commits after evaluation insert and completion
- each composed case uses the real writer path and fresh readback assertions for one observation/artifact/evaluation/input, four WER values with one primary, and a succeeded evaluation
- independent Tier 3 review: PASS, no findings

## Known limitations / rollout
- this host does not have `pg_ctl`, so CI is the first runtime execution of the new PostgreSQL-composed cases
- exhausted retries can still leave a normalized gap; eliminating that residual requires an outbox or reconciliation mechanism and is outside this PR
- after deployment, restart the seven-day clean-observation window before deciding on read cutover
- this PR does not enable S2S, run a backfill, switch normalized reads, or deploy production changes

Linear: https://linear.app/coval/issue/BENCH-797/prevent-normalized-stttts-dual-write-gaps

### DIFF
--- a/runner/src/coval_bench/db/writer.py
+++ b/runner/src/coval_bench/db/writer.py
@@ -71,6 +71,38 @@ class RunWriter:
     ) -> None:
         self._pool = pool
 
+    def pool_diagnostics(self) -> dict[str, int]:
+        """Return a stable, read-only snapshot of pool health counters."""
+        keys = (
+            "pool_min",
+            "pool_max",
+            "pool_size",
+            "pool_available",
+            "requests_waiting",
+            "requests_num",
+            "requests_queued",
+            "requests_wait_ms",
+            "requests_errors",
+            "usage_ms",
+            "connections_num",
+            "connections_ms",
+            "connections_errors",
+            "connections_lost",
+            "returns_bad",
+            "pool_timeout_ms",
+        )
+        diagnostics = {key: 0 for key in keys}
+        stats = self._pool.get_stats()
+        for key in keys:
+            if key in stats:
+                diagnostics[key] = int(stats[key])
+        diagnostics["pool_min"] = int(self._pool.min_size)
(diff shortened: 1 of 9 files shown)


ben-coval (changes requested):
I think we should remove the persistence semaphore. The connection pool already bounds concurrency, and the observed PoolTimeout burst coincided with the migration deadlock addressed by #580/#582. We should keep the retries and blank-error normalization though.

ben-coval (approved):

PR #594, opened 2026-09-03
[BENCH-766] Model pricing in Postgres: admin-recorded rates with history

cooperr1:
Linear: [BENCH-766](https://linear.app/coval/issue/BENCH-766/open-model-pricing-registry-on-get-v1pricing-static-no-db)

## What

- `benchmarks_v2.pricing_rates` (migration `20260903_0027`): append-only log, seeded with the 61 rates verified against providers' public pricing pages. Nothing updates or deletes a row; a correction appends and marks the earlier row superseded, a future date schedules a change, null unit + price records "no known public rate".
- `GET /v1/pricing?as_of=`: the rate in force on a day (default today) with earlier spans as `history`. Same roster and embargo filters as every other data endpoint.
- `GET` / `POST /v1/admin/pricing`: coval-org only via the existing `require_coval_admin`. Each recording is stamped with the caller's Clerk user id and email. Writes pass the ratesheet `PricingEntry` rules.
- Downgrade refuses to drop the table once admin recordings exist.

## Deploy

1. Merge. After the release deploys, run the migration the usual way:
   `gcloud run jobs execute benchmarks-runner --region=us-east1 --project=coval-benchmarks-prod --args="db,migrate"`
2. Grant: coval-ai/benchmark-infra#169 gives the api role SELECT + INSERT on `pricing_rates`, applied via Atlantis after the table exists.
3. Then the web PR: coval-ai/benchmarks-web#76.

Until the grant lands, `GET /v1/pricing` returns 503 and admin writes fail closed. Nothing else is affected.

## Verification

`cd runner && uv run ruff check . && uv run ruff format --check . && uv run mypy --strict src tests && uv run pytest -q --disable-socket --allow-unix-socket --allow-hosts=127.0.0.1,::1` — 2010 passed. Live on the local stack with a mock Clerk issuer: 401 no token, 403 other org, 422 bad unit / float price / bad URL / two years out, 201 stamped recording, repeat 200, public read serves the new rate with the old one in history, `?as_of=yesterday` serves the old one.

### DIFF
--- a/runner/src/coval_bench/api/routers/pricing.py
+++ b/runner/src/coval_bench/api/routers/pricing.py
@@ -1,0 +1,130 @@
+# Copyright 2026 The Coval Benchmarks Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""GET /v1/pricing — the rates in force on a day, with their history.
+
+The response carries one entry per listed model that has ever had a rate
+recorded: the rate in force on ``as_of`` (today by default), the earlier rates
+it succeeded, and the packaged datasets' usage (minutes of audio in,
+characters spoken), so a client can turn a rate into a benchmark-pass cost by
+pure arithmetic — against the datasets the site currently reports, since the
+usage list also keeps retired datasets whose artifacts still reference them.
+
+Rates live in ``benchmarks_v2.pricing_rates`` and change through the admin
+API (see ``admin_pricing.py``), never by deploy; ``registries.pricing.resolve``
+is the one rule for which recording describes which day. A rate scheduled for
+a future day prices nothing until then, and ``as_of`` may not be in the future
+either — the site never shows a price that is not yet in force.
+
+Visibility follows the model roster: a rate serves only while the roster
+lists its model — uncollected models included, since the site shows them
+greyed out — and unpublished models are hidden unless the caller's bearer
+token clears them, the same embargo every other data endpoint applies.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from datetime import UTC, date, datetime
+from typing import Any
+
+import structlog
+from fastapi import APIRouter, Depends, HTTPException, Query
+from posthog import Posthog
+from psycopg_pool import AsyncConnectionPool
+from starlette.requests import Request
+
+from coval_bench.api.deps import capture_api_event, get_models, get_pool, get_posthog
+from coval_bench.api.internal import hidden_early_access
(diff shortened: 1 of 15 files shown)


runner/src/coval_bench/api/routers/pricing.py
```diff
@@ -0,0 +1,130 @@
...
+    try:
+        recordings = await PricingStore(pool).recordings()
+    except Exception as exc:
+        logger.error("pricing_log_unavailable", exc_info=True)
+        raise HTTPException(503, "the pricing log is unavailable") from exc
+
+    roster = {(m.benchmark, m.provider, m.model) for m in models}
+    rates: list[PricingRateOut] = []
+    for key, timeline in sorted(timelines(recordings).items()):
+        if key not in roster or (key[1], key[2]) in hidden:
```
coval-cale, line 114: [P2] Keep pricing attached to an immutable model identity. This joins the pricing log to the current roster using the mutable `(benchmark, provider, model)` key. After a model is renamed, its old key becomes available to `POST /admin/models`; a different model later created under that key will silently inherit the original model's entire pricing timeline, potentially including embargoed rates. Please store an immutable model ID, migrate rates during rename, or reserve historical keys so they cannot be reused.

runner/src/coval_bench/db/migrations/versions/20260903_0027_pricing_rates.py
```diff
@@ -0,0 +1,733 @@
...
+)
+
+
+def upgrade() -> None:
+    """Create the pricing log and seed it with the packaged rates."""
+    op.execute(
+        """
+        CREATE TABLE benchmarks_v2.pricing_rates (
+            id                  BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
+            benchmark           TEXT NOT NULL CHECK (benchmark IN ('STT', 'TTS', 'S2S')),
```
coval-cale, line 672: [P2] Reject unsupported modalities before inserting. `AdminRateCreate` accepts the full `Benchmark` enum, and an unpriced LLM entry skips `PricingEntry`'s unit validation. For a registered LLM model, `model_exists()` therefore succeeds and this CHECK rejects the insert, surfacing an uncaught database error as a 500. Please restrict the admin schema to supported pricing modalities or make the database and validation rules agree.

coval-cale (approved):
Approving overall. I left two follow-up correctness comments around model identity across renames and unsupported benchmark validation.

coval-cale (approved):

PR #605, opened 2026-09-08
Proxy and sync every LLM provider through one definition

coval-cale:
The proxy, Coval agent, sync command, and fetch specs all hardcoded Phonely. Route turns at /llm/{provider}, derive the Coval agent names and endpoints from the provider slug, and drive sync-llm and the LLM fetch specs from one provider-to-model mapping. Phonely's generated names and paths match its live values, so this deploys as a no-op.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ coverage.xml
 # IDE / editor
 .idea/
 .vscode/
+.zed/
 *.swp
 *.swo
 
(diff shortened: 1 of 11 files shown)


seribaymadina:
**Findings**

- **P1:** `LLM_MODELS` now controls accepted routes, but app startup still initializes only `PhonelyClient`. Adding another provider will create/sync an agent whose `/llm/{provider}` endpoint always returns `503`. Keep the provider registry and client initialization in sync.
- **P1:** `simulation_id` is caller-controlled and used directly for timing inserts. A caller with the proxy secret can attribute turns to another simulation. Validate ownership/existence or derive the simulation identity server-side.

I found no obvious new auth bypass, SSRF, or secret-leak regression.

seribaymadina (approved):
NOICE! left couple comments

PR #617, opened 2026-09-09
[BENCH-846] Make pooled WER the headline with a mean fallback

coval-cale:
On the normalized path avg_value for WER is the pooled ratio when every clip in the window carries counts and the per-clip mean otherwise, with the mean always on mean_value; the S/I/D split and series value follow the same rule. Series points expose error_sum and reference_word_sum so consumers can pool across buckets. /v1/leaderboard gains a normalized path behind the reads flag and ranks on the same headline. Methodology doc gains the aggregation section, a dated not-comparable note, and ADR-024.

### DIFF
--- a/docs/methodology.md
+++ b/docs/methodology.md
@@ -49,7 +49,7 @@ retired manifests (`stt-v2.json`) stay for historical reproducibility.
 **Aggregation across datasets.** Each dataset runs as its own run (one
 `DATASET_ID` per execution), and a result's dataset is derived from its parent
 run. The aggregation layer pools every result in the window regardless of
-dataset, so headline stats (e.g. average WER) blend all datasets that ran.
+dataset, so headline stats (e.g. pooled WER) blend all datasets that ran.
 `/v1/results` exposes each row's `dataset_id` and takes a `dataset` filter for
 per-dataset inspection. See ADR-023.
 
@@ -103,6 +103,27 @@ with `EnglishTextNormalizer` per [ADR-021](#adr-references). WER values
 before and after the change are **not comparable**. Because result rows do not
 carry a normalization-version marker, historical rows cannot be reliably
 segmented by normalization pipeline version.
+
+### Aggregation
+
+The headline WER for a model is corpus-level: the sum of substitutions,
+deletions, and insertions across every clip in the window, divided by the sum
+of reference words. Each per-clip row carries those four counts, so the figure
+is exact rather than an average of ratios. The per-clip distribution (median,
+percentiles, spread) is still reported from the per-clip ratios.
+
+The two disagree by design. Reference clips run from 4 to 41 words, so a mean
+of per-clip ratios lets one miss on a 4-word clip weigh the same as a perfect
+41-word clip, and the empty-reference convention above turns a clip with no
+reference into a raw insertion count. Pooling weights every word equally and
+absorbs both.
+
+**Methodology change (2026-09).** WER was previously the mean of per-clip
+ratios. Counts are recorded from this point forward and were not backfilled,
+so a window reports the pooled figure only once every clip in it carries
+counts; until then it falls back to the mean and says so (`pooled_value` is
+null, `mean_value` is always present). Headline numbers before and after are
+**not comparable**. See [ADR-024](#adr-references).
 
 ## 4. Latency metrics (TTFA)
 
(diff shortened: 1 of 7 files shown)


runner/src/coval_bench/api/routers/leaderboard.py
```diff
@@ -113,13 +117,26 @@ async def get_leaderboard(
         "metric": metric,
         "benchmark": benchmark,
         "dataset": _PRIMARY_DATASET_BY_BENCHMARK.get(benchmark, DATASET_ALL),
+        "interval": WINDOW_INTERVALS[window],
     }
-    sql = _MV_SQL_TEMPLATE.format(view=WINDOW_VIEWS[window])
+    normalized = settings.normalized_dashboard_reads_enabled
+    sql = (
+        _NORMALIZED_STATS_SQL if normalized else _MV_SQL_TEMPLATE.format(view=WINDOW_VIEWS[window])
```
kdmelon, line 125: **[P1] Keep LLM leaderboard reads on legacy storage during cutover.** This flag now redirects `TTFT+LLM` to normalized tables, but `s2s/fetch_v2v.py` still persists LLM results only through `record_results`; its normalized dual-write branch explicitly requires `Benchmark.S2S`. Neither backfill includes LLM, and the readiness checker excludes it too. A PostgreSQL-backed HTTP probe returned the existing Phonely entry with the flag off and `entries: []` with it on. Preserve the legacy LLM read path until its writer and historical coverage are ready, and cover the flag transition in the LLM leaderboard test.

runner/src/coval_bench/api/routers/aggregates.py
```diff
@@ -273,11 +283,10 @@
  MAX(max_value) FILTER (WHERE value_key = 'primary') AS max_value,
  MAX(value_sum) FILTER (WHERE value_key = 'primary') AS value_sum,
  MAX(sample_count) FILTER (WHERE value_key = 'primary') AS sample_count,
- CASE WHEN COUNT(*) = 5 AND MIN(sample_count) = MAX(sample_count)
-      THEN 100 * SUM(value_sum) FILTER (WHERE value_key IN
-             ('substitution_count', 'deletion_count', 'insertion_count'))
-           / NULLIF(SUM(value_sum) FILTER (WHERE value_key = 'reference_words'), 0)
-      END AS pooled_value
+ CASE WHEN {_BUCKET_COUNTS_COMPLETE} THEN {_BUCKET_ERROR_SUM} END AS error_sum,
+ CASE WHEN {_BUCKET_COUNTS_COMPLETE} THEN {_BUCKET_REFERENCE_SUM} END AS reference_word_sum,
```
kdmelon, line 289: **[P2] Forward both count sums through the 30-day compact query.** These fields are populated here, but `_COMPACT_SERIES_TAIL`'s final SELECT omits them. The 30-day aggregate response therefore defaults both to null even for fully counted buckets. Reproduced with the same bucket across all windows: 24h/7d return `error_sum=1`, `reference_word_sum=40`, and `pooled_value=2.5`; 30d retains `pooled_value=2.5` but returns null count sums. Consumers cannot pool the returned buckets using the new contract. Include both columns in the compact projection, adding null aliases to the legacy base query if the tail remains shared.

docs/methodology.md
```diff
@@ -104,6 +104,27 @@ before and after the change are **not comparable**. Because result rows do not
...
+percentiles, spread) is still reported from the per-clip ratios.
+
+The two disagree by design. Reference clips run from 4 to 41 words, so a mean
+of per-clip ratios lets one miss on a 4-word clip weigh the same as a perfect
+41-word clip, and the empty-reference convention above turns a clip with no
+reference into a raw insertion count. Pooling weights every word equally and
+absorbs both.
+
+**Methodology change (2026-09).** WER was previously the mean of per-clip
+ratios. Counts are recorded from this point forward and were not backfilled,
```
kdmelon, line 122: **Related table-cutover blocker inherited from #612:** the new count fields are rejected by the current live-row backfill validator. `migrations/backfill_normalized_storage.py:_values` reconstructs only primary WER plus the three percentage components, while `_live_validation_result` requires exact equality with every stored metric value. A deterministic probe using the actual normalized writer accepts a healthy WER=6% observation without counts, then rejects the identical observation with valid S=3, D=2, I=1, N=100 counts as `category='value'`. This predates #617 and can be handled in a separate cutover fix: validate the known optional all-or-none count group while retaining exact checks on the legacy public values and unrelated fields. The mean fallback documented here means historical count backfill itself is not required.

kdmelon (approved):
Reviewed together with the normalized-table cutover at d90e335e. The pooled WER calculation and mixed-history mean fallback check out. All 57 existing aggregate/leaderboard API tests, 22 readiness tests, and changed-file Ruff checks passed. Two targeted PostgreSQL-backed API probes reproduced the issues documented inline; a separate deterministic probe confirmed the inherited #612 validator mismatch. Production cutover readiness was not revalidated.

PR #619, opened 2026-09-09
[BENCH-861] Generic OpenAI-compatible client for the LLM proxy

coval-cale:
Prep for adding gpt-4.1 (BENCH-847) and gemini-2.5-flash (BENCH-848) to the LLM benchmark.

- Move `TurnAccumulator` out of `llm/phonely.py` into `llm/openai_compat.py`; it parses the chat-completions SSE format every OpenAI-compatible endpoint emits.
- Add `OpenAICompatClient` for models we prompt ourselves. Takes base URL, key, model, system prompt, tools, and an optional `extra_body` for per-provider switches (thinking off). Sessions are minted locally.
- Add `llm/dental.py`, which loads the dental contract prompt and the five tool definitions once in OpenAI function-tool shape.
- Phonely keeps its own client since it hosts the agent and owns a real session.

No new providers in this PR. Each new model becomes a `CLIENT_FACTORIES` entry plus its settings fields.

Linear: https://linear.app/coval/issue/BENCH-861

### DIFF
--- a/runner/src/coval_bench/llm/dental.py
+++ b/runner/src/coval_bench/llm/dental.py
@@ -1,0 +1,32 @@
+# Copyright 2026 The Coval Benchmarks Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""The dental agent as a prompt and tool list, for models we drive ourselves."""
+
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass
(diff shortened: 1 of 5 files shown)


cooperr1 (approved):

kdmelon (approved):
🦷

PR #625, opened 2026-09-10
Accept a composite judge's numeric fraction in the instruction mapper

seribaymadina:
Traced real Coval data while checking alignment with a report Brooke's running: the three domain-specific instruction-following judges (Healthcare/Home Service/Customer Service) report literal `YES`/`NO`/`UNKNOWN` strings per conversation, which the existing mapper already handles correctly. Expected Behavior Adherence reports something different — its own already-computed `0-1` "percentage of criteria met" as a numeric value, through the same metric id shape. The mapper only understood the string verdict, so a numeric value raised `InvalidInstructionVerdict` on every conversation and silently discarded the whole metric for that run — as shipped in #623, wiring the industry pipeline to EBA would have ingested zero rows, with only a warning log to show for it.

Numeric raw values are now scaled directly to a percentage (`0.75` -> `75.0`) instead of being run through verdict classification; string values still go through the existing YES/NO/UNKNOWN path unchanged. Added two tests exercising the numeric branch directly and through `_s2s_rows`.

### DIFF
--- a/runner/src/coval_bench/s2s/fetch_v2v.py
+++ b/runner/src/coval_bench/s2s/fetch_v2v.py
@@ -490,8 +490,16 @@ def _population_mismatch(
 def _instruction_value(raw: object) -> tuple[float | None, ResultStatus] | None:
     """YES -> 100.0, NO -> 0.0, UNKNOWN -> no row, so the mean is YES / (YES + NO).
 
-    Raises InvalidInstructionVerdict on any value outside the contract.
+    A numeric raw value is a composite judge's own already-computed fraction of
+    criteria met (e.g. Expected Behavior Adherence's 0-1 "percentage_of_criteria_met"),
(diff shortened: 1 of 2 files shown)


seribaymadina:
Superseded by #624, which solved the same problem (EBA reports a numeric 0-1 fraction, not a YES/NO verdict) with a cleaner design: a dedicated Metric.EXPECTED_BEHAVIOR_ADHERENCE + _expected_behavior_value mapper fetched via its own expected_behavior_metric_id_attr, alongside the domain judge rather than overloading _instruction_value. Verified directly against post-#624 main: _instruction_value is back to strict binary-only (good, preserves its contract), and this branch now conflicts with main. Closing rather than merging — no live benefit, and it would weaken a mapper whose only real callers are confirmed YES/NO/UNKNOWN strings.

PR #626, opened 2026-09-10
[BENCH-868] Pass LLM Coval agent ids from sync to fetch directly

coval-cale:
`AgentSpec` now carries the Coval agent id rather than the name of a Settings field that holds it. Both benchmarks build the same shape and the fetch reads `spec.agent_id` everywhere.

- S2S: the module-level table becomes `s2s_specs(settings)`, with each entry reading its id as a typed field (`settings.coval_s2s_openai_agent_id`). Same env vars, same behavior; a misspelled field now fails mypy instead of skipping an agent at runtime.
- LLM: `llm_specs(models, agent_ids)` takes the provider-to-id map the sync reconciled, passed through `_run_fetch` / `fetch_and_write_v2v` as `llm_agent_ids`. The three `coval_llm_*_agent_id` settings and the `model_copy(update=...)` hand-off are gone.

No runtime change: an unset id still warns and skips. Adding an LLM provider no longer needs a settings field.

Follow-up in benchmark-infra: drop the unused `coval_llm_phonely_agent_id` variable.

Makes sure that LLM and S2S models have the same AgentSpec shape. Makes it easy to, down the line, move the S2S model info into the db.

### DIFF
--- a/runner/src/coval_bench/config.py
+++ b/runner/src/coval_bench/config.py
@@ -190,10 +190,6 @@ class Settings(BaseSettings):
     coval_s2s_xai_think_fast_2_agent_id: str | None = None
     coval_s2s_gray_agent_id: str | None = None
     coval_s2s_red_agent_id: str | None = None
-    # Coval agent id for the Phonely text agent (opaque, not secret); unset skips it.
-    coval_llm_phonely_agent_id: str | None = None
-    coval_llm_openai_agent_id: str | None = None
-    coval_llm_google_agent_id: str | None = None
     # The S2S instruction-adherence metric id (opaque, not secret). Optional: the
     # fetch pulls its per-conversation scores only when set, so latency still
     # ingests without it.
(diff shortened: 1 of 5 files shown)


callumreid (approved):
Verdict: APPROVE. Reviewed base main at f355828ac3012fab5897036ac9f28c932310251b against head 9280c019af06a54da4088596a5ce9c65e9404d1b for the AgentSpec/LLM agent-id handoff changes. Exact-head CI is green: CodeQL, Switchboard Plan checks, Benchmarks Runner CI Adapter / Runner CI, Switchboard Gate. Local validation passed: focused pytest, ruff, and mypy over the changed source/tests. Quality gates checked for correctness, regressions, test coverage, architecture/scope, error handling, performance, comments, migrations, and PR-description accuracy; no blocking defects found. Residual risk: low, limited to not exercising live Coval API/database ingestion in review mode.

seribaymadina (approved):
very cool!
couple comments if im understanding things correctly and feel free to  push back: 
- This simplifies the LLM flow because sync already knows the agent IDs. S2S still needs settings, Terraform wiring, and a Python entry for each new agent. Can we track removing that remaining setup as a follow-up?
- When we move S2S configuration into the database, we should include the workspace, test set, and metric IDs alongside the agent ID. Otherwise, adding an agent could still require settings and infrastructure changes.
- For S2S, we’ll need separate records for each model and industry configuration. A provider-to-ID dictionary alone won’t work because the same provider can have several different agents.

PR #627, opened 2026-09-10
[BENCH-872] Add the tts-v2 manifest

coval-cale:
Ship the 330-item TTS v2 bank as tts-v2.json next to the untouched tts-v1. TTS manifest and dataset items gain optional metadata and a spoken reference, which the orchestrator now uses as the WER reference when present. Nothing loads tts-v2 in production yet.


kdmelon (changes requested):
close the pr, ill regen the dataset

PR #630, opened 2026-09-10
[BENCH-872] Load the private tts-v2 manifest from GCS and score WER against the spoken reference

coval-cale:
## What

- `tts-v2.json` ships as a pointer only: bucket, object path, and sha256 of the manifest.
- `Manifest` gains an optional `remote` block; a file may carry items or a pointer, never both.
- The loader resolves a pointer by fetching the manifest from `coval-benchmarks-datasets-private` (coval-ai/benchmark-infra#188), verifying the hash, and caching it. The download-and-verify step is shared with STT audio. Private fetches use real credentials rather than the anonymous public-bucket client.
- `TTSManifestItem` / `TTSDatasetItem` gain optional `spoken_reference`, `kind`, `vertical`, `difficulty`, `family`, `length_bucket`, `tags`, and `phonetic`. tts-v1 items are unchanged.
- The orchestrator scores WER against `spoken_reference` when present, falling back to `transcript`.

## Why

The TTS v2 authoring bank (BENCH-759) is proprietary.

## Reviewer notes

- Nothing loads tts-v2 in production yet; the orchestrator still hardcodes tts-v1. BENCH-873 makes the dataset id a setting and adds stratified sampling.
- The manifest is uploaded and its hash matches the pointer. I loaded it end to end against the real bucket with the runner code.
- tts-v2 rows keep their own dataset id, so the existing per-dataset aggregates apply. Do not add a cross-dataset TTS blend: v1 scores against normalized written text, v2 against an authored spoken form, and the two are not comparable.
- `tts-v1.json` is untouched; its byte hash ties historical runs to their items.

Linear: https://linear.app/coval/issue/BENCH-872

### DIFF
--- a/runner/src/coval_bench/datasets/loader.py
+++ b/runner/src/coval_bench/datasets/loader.py
@@ -136,11 +136,8 @@ class Dataset(BaseModel):
     items: list[DatasetItem]
 
 
-class TTSDatasetItem(BaseModel):
+class TTSDatasetItem(TTSManifestItem):
     """A single TTS benchmark item (text-only, no audio to fetch)."""
-
-    testcase_id: str = Field(min_length=1)
-    transcript: str
 
 
 class TTSDataset(BaseModel):
@@ -256,6 +253,34 @@ def _fetch_blob(
     blob.download_to_filename(str(dest))
 
 
+def _fetch_verified(
+    *,
+    client: storage.Client,
+    bucket: str,
+    object_path: str,
+    sha256: str,
+    local_path: Path,
+) -> Path:
+    """Return *local_path* holding *object_path*, downloading from GCS if needed.
+
+    If the cached file already has the correct SHA256, the download is skipped.
+    On SHA256 mismatch after download, raises :class:`DatasetIntegrityError`.
+    """
+    if local_path.exists():
+        if _sha256_file(local_path) == sha256:
+            logger.debug("cache_hit", path=object_path)
+            return local_path
+        logger.warning("cached_file_hash_mismatch", path=str(local_path))
+
+    logger.info("fetching_dataset_object", bucket=bucket, object=object_path)
+    _fetch_blob(client, bucket, object_path, local_path)
+
+    actual = _sha256_file(local_path)
+    if actual != sha256:
+        raise DatasetIntegrityError(object_path, sha256, actual)
+    return local_path
+
+
 def _fetch_and_verify(
     *,
     client: storage.Client,
@@ -264,31 +289,42 @@ def _fetch_and_verify(
     item: STTManifestItem,
     cache_dir: Path,
 ) -> Path:
-    """Return local path for *item*, downloading from GCS if needed.
+    """Return local path for *item*, downloading from GCS if needed."""
+    return _fetch_verified(
+        client=client,
+        bucket=bucket,
+        object_path=f"{dataset_id}/{item.path}",
+        sha256=item.sha256,
+        local_path=cache_dir / dataset_id / item.path,
+    )
 
-    If the cached file already has the correct SHA256, the download is skipped.
-    On SHA256 mismatch after download, raises :class:`DatasetIntegrityError`.
-    """
-    local_path = cache_dir / dataset_id / item.path
 
-    # Cache hit check
-    if local_path.exists():
-        if _sha256_file(local_path) == item.sha256:
-            logger.debug("cache_hit", path=str(item.path))
-            return local_path
-        logger.warning("cached_file_hash_mismatch", path=str(local_path))
-
-    # Download
-    gcs_object = f"{dataset_id}/{item.path}"
-    logger.info("fetching_dataset_object", bucket=bucket, object=gcs_object)
-    _fetch_blob(client, bucket, gcs_object, local_path)
-
-    # Post-download integrity check
-    actual = _sha256_file(local_path)
-    if actual != item.sha256:
-        raise DatasetIntegrityError(item.path, item.sha256, actual)
-
-    return local_path
+def _resolve_manifest(
+    dataset_id: str,
+    *,
+    cache_dir: Path | None,
+    storage_client: storage.Client | None,
+) -> Manifest:
+    """Load the packaged manifest, following a private ``remote`` pointer if present."""
+    packaged = _load_manifest(dataset_id)
+    if packaged.remote is None:
+        return packaged
+    # A private bucket needs real credentials; the anonymous public-bucket client cannot read it.
+    client = storage_client if storage_client is not None else storage.Client()
+    resolved_cache = cache_dir if cache_dir is not None else _default_cache_dir()
+    local_path = _fetch_verified(
+        client=client,
+        bucket=packaged.remote.bucket,
+        object_path=packaged.remote.path,
+        sha256=packaged.remote.sha256,
+        local_path=resolved_cache / dataset_id / "manifest.json",
+    )
+    manifest = Manifest.model_validate_json(local_path.read_text(encoding="utf-8"))
+    if manifest.id != dataset_id or manifest.remote is not None:
+        raise ManifestAlignmentError(
+            f"remote manifest for '{dataset_id}' is not a full manifest with that id"
+        )
+    return manifest
 
 
 # ---------------------------------------------------------------------------
@@ -327,7 +363,7 @@ def load_stt_dataset(
     Dataset
         Fully-verified dataset with local file paths.
     """
-    manifest = _load_manifest(dataset_id)
+    manifest = _resolve_manifest(dataset_id, cache_dir=cache_dir, storage_client=storage_client)
     if dataset_id in WILDASR_ENV_FAMILY:
         _assert_family_alignment(dataset_id, manifest)
     resolved_cache = cache_dir if cache_dir is not None else _default_cache_dir()
@@ -371,24 +407,24 @@ def load_stt_dataset(
 def load_tts_dataset(
     dataset_id: str,
     *,
-    settings: Settings,  # noqa: ARG001 – kept for uniform call signature
-    cache_dir: Path | None = None,  # noqa: ARG001 – no files to cache for TTS
-    storage_client: storage.Client | None = None,  # noqa: ARG001 – no GCS needed
+    settings: Settings,
+    cache_dir: Path | None = None,
+    storage_client: storage.Client | None = None,
     sample_size: int | None = None,
     rng: random.Random | None = None,
 ) -> TTSDataset:
-    """Load a TTS dataset (text prompts only; no GCS fetch required).
+    """Load a TTS dataset (text prompts only; GCS is touched only for a private manifest).
 
     Parameters
     ----------
     dataset_id:
         Manifest identifier, e.g. ``"tts-v1"``.
     settings:
-        Accepted for call-signature symmetry; not used for TTS.
+        Runner settings; used to build a GCS client for private manifests.
     cache_dir:
-        Accepted for call-signature symmetry; not used for TTS.
+        Override the local cache directory for a fetched private manifest.
     storage_client:
-        Accepted for call-signature symmetry; not used for TTS.
+        Injectable GCS client (for tests; production passes ``None``).
     sample_size:
         Draw this many prompts at random; ``None`` uses all.
     rng:
@@ -399,7 +435,7 @@ def load_tts_dataset(
     TTSDataset
         Dataset with text-only items.
     """
-    manifest = _load_manifest(dataset_id)
+    manifest = _resolve_manifest(dataset_id, cache_dir=cache_dir, storage_client=storage_client)
 
     tts_items: list[TTSDatasetItem] = []
     for raw_item in _sample_items(list(manifest.items), sample_size, rng):
@@ -407,12 +443,7 @@ def load_tts_dataset(
             raise TypeError(
                 f"Dataset '{dataset_id}' contains non-TTS items; use load_stt_dataset instead."
             )
-        tts_items.append(
-            TTSDatasetItem(
-                testcase_id=raw_item.testcase_id,
-                transcript=raw_item.transcript,
-            )
-        )
+        tts_items.append(TTSDatasetItem.model_validate(raw_item.model_dump()))
 
     return TTSDataset(id=manifest.id, version=manifest.version, items=tts_items)
 
@@ -431,7 +462,7 @@ def load_dataset(
     Prefer the explicit ``load_stt_dataset`` / ``load_tts_dataset`` functions
     when the caller knows the dataset type ahead of time.
     """
-    manifest = _load_manifest(dataset_id)
+    manifest = _resolve_manifest(dataset_id, cache_dir=cache_dir, storage_client=storage_client)
     if manifest.items and isinstance(manifest.items[0], STTManifestItem):
         return load_stt_dataset(
             dataset_id,
--- a/runner/src/coval_bench/datasets/loader.py
+++ b/runner/src/coval_bench/datasets/loader.py
@@ -420,7 +420,8 @@ def load_tts_dataset(
     dataset_id:
         Manifest identifier, e.g. ``"tts-v1"``.
     settings:
-        Runner settings; used to build a GCS client for private manifests.
+        Accepted for call-signature symmetry; not consulted. The private-manifest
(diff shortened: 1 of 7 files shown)


runner/src/coval_bench/runner/orchestrator.py
```diff
@@ -429,9 +429,6 @@ async def _run_stt_item(
             kwargs["project_id"] = settings.google_project_id
         elif entry.provider == "baseten":
             kwargs["ws_url"] = _get_baseten_stt_url()(settings, entry.model)
-        elif entry.provider == "guava":
```
callumreid, line 432: **P2 — Guava STT runs now fail before contacting the provider.** Removing these Guava-specific kwargs means `_run_stt_item` builds only `{api_key, model}` for `entry.provider == "guava"`, but `GuavaSTTProvider.__init__` still requires `base_url` and derives the websocket URL from it. I verified this path with an in-process `_run_stt_item` call using `Settings(guava_api_key=..., guava_base_url=..., guava_stt_domain=...)`: it returns failed metric rows with `guava_base_url is required for the Guava STT provider`. The focused tests passed because they exercise the CLI smoke path, which still passes these kwargs, not the orchestrator path. Please restore the `base_url` and `domain` wiring here (and ideally add an orchestrator-level Guava regression test) so the newly registered Guava STT model does not report every benchmark item as a configuration failure.

callumreid (dismissed):
Verdict: NOT READY. Reviewed benchmarks#630 against current origin/main at exact head b5108861888ed3a36204fd0156e9bbbf5816db66. The TTS-v2 private manifest flow and spoken_reference WER path are covered by the added tests and passed focused validation, but the diff also regresses Guava STT orchestration. Validation: plain pytest failed because coval_bench was not installed on the ambient Python path; uv run pytest tests/unit/test_dataset_loader.py tests/unit/test_orchestrator.py::test_tts_wer_scores_against_spoken_reference tests/providers/stt/test_guava.py::test_guava_smoke_passes_configured_domain passed (39 passed); uv run ruff check src/coval_bench/datasets/loader.py src/coval_bench/datasets/manifest.py src/coval_bench/runner/orchestrator.py tests/unit/test_dataset_loader.py tests/unit/test_orchestrator.py passed; an in-process _run_stt_item reproducer for provider=guava with configured guava_base_url returned failed rows with 'guava_base_url is required'. Residual risk: did not access the private GCS bucket, so the uploaded object/hash claim is from the PR body and unit tests rather than live bucket verification.

runner/src/coval_bench/runner/orchestrator.py
coval-cale: Insane catch! Nice bro!

coval-cale: Good catch, thanks. Root cause: I copied orchestrator.py wholesale from the closed PR's commit, which predated #615, so the Guava kwargs were reverted along the way. Rebuilt the file from main with only the spoken-reference WER line changed, and added an orchestrator-level test that pins base_url and domain reaching the Guava constructor. a86f1c0c.

kdmelon:
**P2 — Preserve immutable remote manifest revisions.**

The update command in [README.md:375–376](https://github.com/coval-ai/benchmarks/blob/a86f1c0c7d4e1f0be1d6ce1842b8f5877b30f40d/runner/src/coval_bench/datasets/manifests/README.md#L375-L376) overwrites `datasets/tts-v2/manifest.json` before repinning its SHA. Older runner versions retain the previous SHA, but the loader fetches the current object at that path. After a bank update, historical reruns and rollbacks therefore fail with `DatasetIntegrityError` on an empty cache.

I reproduced this through `load_tts_dataset` with a fake bucket: the initial load succeeds; replacing the remote bytes leaves warm-cache loads working, while the same pinned runner with a fresh cache fails hash verification. Please upload each revision under an immutable version/hash path and update `remote.path`, or pin and load a specific GCS generation.

Reviewed `a86f1c0c`. All 124 focused dataset-loader/orchestrator tests passed, along with Ruff lint and formatting checks. A composed synthetic probe verified manifest loading/cache, metadata preservation, spoken-reference scoring, and actual `RunWriter` SQL serialization. Live GCS, providers, and PostgreSQL were not exercised. The earlier Guava regression is fixed on this head.

coval-cale:
Fixed in fe94b1f7. The object name is now the manifest's sha256 (datasets/tts-v2/<sha>.json), so a bank update is a new object and a pinned runner with a cold cache resolves the bytes it was released with. RemoteManifest rejects a path that does not contain its sha, and the README update flow uploads to the hashed name. The current manifest is uploaded under the new name; the old fixed-path object is left in place and nothing references it.

runner/src/coval_bench/datasets/loader.py
```diff
@@ -371,24 +407,24 @@ def load_stt_dataset(
...
-    """Load a TTS dataset (text prompts only; no GCS fetch required).
+    """Load a TTS dataset (text prompts only; GCS is touched only for a private manifest).
 
     Parameters
     ----------
     dataset_id:
         Manifest identifier, e.g. ``"tts-v1"``.
     settings:
-        Accepted for call-signature symmetry; not used for TTS.
+        Runner settings; used to build a GCS client for private manifests.
```
callumreid: **P3 — the docstring claims `settings` builds the private-manifest GCS client, but the TTS loader never consults it.** `_resolve_manifest` builds the client from the injected `storage_client` or ambient ADC (`loader.py:313`: `client = storage_client if storage_client is not None else storage.Client()`); `load_tts_dataset` never passes `settings` anywhere, so `google_application_credentials` and every other settings field have no effect on the private fetch. The hunk also drops the old `# noqa: ARG001 – kept for uniform call signature` comment for `settings`, which reads as "now used". Focused tests pass either way because they inject `storage_client`, so nothing catches the doc drift. Smallest fix:

```suggestion
        Accepted for call-signature symmetry; not consulted — the private-manifest
        client comes from ``storage_client`` or application-default credentials.
```

runner/src/coval_bench/datasets/manifest.py
```diff
@@ -62,11 +109,14 @@ class Manifest(BaseModel):
...
-    items: list[STTManifestItem | TTSManifestItem]
+    items: list[STTManifestItem | TTSManifestItem] = Field(default_factory=list)
+    remote: RemoteManifest | None = None
 
     @model_validator(mode="after")
     def items_consistent(self) -> Manifest:
         """Ensure items are homogeneous and have distinct effective identities."""
+        if self.items and self.remote is not None:
+            raise ValueError(f"Manifest '{self.id}' cannot carry both items and a remote pointer")
         if not self.items:
```
callumreid, line 122: **P3 — a manifest with neither `items` nor `remote` now validates silently, where it used to fail at load.** On main `items` was a required field, so a packaged manifest that lost its `items` key failed `Manifest.model_validate_json` with `Field required`. After this change (`items` defaults to `[]` to let pointer files carry no items), the same damaged file validates, the dataset loads empty, and the failure only surfaces at run level as the generic `RunStatus.FAILED` / "benchmark produced no results (no providers configured or dataset empty)" (`orchestrator.py:1457`, `orchestrator.py:276`) instead of a precise load-time error. No test covers the items-less/remote-less shape (the new tests cover pointer-with-remote and items-without-remote only). Enforce exactly-one-of in the existing validator:

```suggestion
        """Ensure items are homogeneous and have distinct effective identities."""
        if not self.items and self.remote is None:
            raise ValueError(f"Manifest '{self.id}' must carry either items or a remote pointer")
        if self.items and self.remote is not None:
            raise ValueError(f"Manifest '{self.id}' cannot carry both items and a remote pointer")
```

runner/src/coval_bench/datasets/manifest.py
```diff
@@ -39,13 +41,55 @@ def _sample_id_not_empty(cls, value: str | None) -> str | None:
...
+
+    bucket: str = Field(min_length=1)
+    path: str = Field(min_length=1)
+    sha256: str = Field(pattern=r"^[a-f0-9]{64}$")
+
+    @model_validator(mode="after")
+    def path_is_content_addressed(self) -> RemoteManifest:
+        """A bank update is a new object, so a pinned runner never sees overwritten bytes."""
+        if self.sha256 not in self.path:
+            raise ValueError(f"remote path {self.path!r} must contain its sha256")
```
callumreid, line 91: **P3 — the new integrity guards have no tests, so a future refactor could drop one silently.** `path_is_content_addressed` (here), the items/remote mutual exclusion (`manifest.py:119-120`), and the resolved-manifest alignment check in `_resolve_manifest` (`loader.py:323-326` — wrong `id` or a nested `remote` raises `ManifestAlignmentError`) are all new in this PR and none is exercised: the added loader tests cover fetch/verify/mismatch and pointer shape, and `ManifestAlignmentError` is imported in `test_dataset_loader.py` but never asserted. These guards are what stop a private manifest from pointer-chasing or scoring a different dataset's items. Three small `pytest.raises` cases in `test_dataset_loader.py` (next to the existing duplicate-identity validation tests) would pin them; each fails today if the corresponding guard is deleted, and nothing else would.

callumreid (approved):
**APPROVE** — prior blocker fixed, new work clean; three P3 polish items inline.

Reviewed exact head `fe94b1f7e37f7bfde4f582d2b540bdc0bcb139cd` against fresh `origin/main` (three-dot, 7 files, +312/−41, matching GitHub's counts).

**Delta since my dismissed review** (`b5108861`, 2026-09-10T22:29): the Guava STT regression is **fixed** — the wiring at `orchestrator.py:432-434` is byte-identical to origin/main (`a86f1c0c` rebuilt the file from main rather than the stale copy), and the new `test_stt_guava_receives_base_url_and_domain` pins it by capturing the constructor kwargs (fails without the restore). The new content-addressed private-manifest work verifies clean end to end: packaged pointer → `_fetch_verified` (hash-check cache hit → fetch → post-download sha256 verify, hard `DatasetIntegrityError` on mismatch) → resolved-manifest id/alignment checks (no pointer-chasing) → typed load. WER reference selection (`item.spoken_reference or transcript`) has the correct `compute_wer(reference, hypothesis)` argument order; `tts-v1` production path untouched (orchestrator still hardcodes `tts-v1`, matching the body's BENCH-873 note). No credentials or manifest content logged; anonymous client never used for private fetches.

**Validation:** `uv run pytest` on the loader tests, the spoken-reference WER test, the new Guava pin, and the Guava smoke test — **40 passed**; `uv run ruff check` + `format --check` clean on the changed files; all 5 CI checks green with each run's head_sha verified against this exact head.

**Residual:** Medium on the live-bucket claims (the object exists at the content-addressed path with a matching hash — author-runtime statements, not verifiable offline; the fetch-verify-alignment chain fails hard rather than mis-scoring if they are wrong). High on the code.

runner/src/coval_bench/datasets/loader.py
coval-cale: Took your wording. cc7dc7c9.

runner/src/coval_bench/datasets/manifest.py
coval-cale: Done: the validator now requires items or a remote and still rejects both. The two tests that built empty manifests pinned the shape this forbids, so they are replaced by a parametrized guard test. cc7dc7c9.

runner/src/coval_bench/datasets/manifest.py
coval-cale: Added: a parametrized shape test covers neither/both/non-content-addressed path, and a fetched-manifest test covers wrong id and nested pointer raising ManifestAlignmentError. cc7dc7c9, typing fix in d5a3a802.

runner/src/coval_bench/datasets/loader.py
```diff
@@ -256,41 +253,80 @@ def _fetch_blob(
...
+        bucket=packaged.remote.bucket,
+        object_path=packaged.remote.path,
+        sha256=packaged.remote.sha256,
+        local_path=resolved_cache / dataset_id / "manifest.json",
+    )
+    manifest = Manifest.model_validate_json(local_path.read_text(encoding="utf-8"))
+    if manifest.id != dataset_id or manifest.remote is not None:
+        raise ManifestAlignmentError(
+            f"remote manifest for '{dataset_id}' is not a full manifest with that id"
+        )
```
callumreid, line 326: **P3 — ManifestAlignmentError's class docstring no longer covers all its raise sites.** The class docstring (loader.py:106-111) says it is 'Raised when a WildASR family manifest drifts out of index alignment'; after this PR, _resolve_manifest also raises it for a remote manifest that has the wrong id or a nested pointer — a completely different failure mode. A reader hitting this exception from a private-dataset load will be pointed at WildASR. Tests pass either way because the new alignment tests only assert the exception type. Smallest fix: broaden the class docstring to cover both raise sites (family misalignment and unaligned remote-manifest resolution).

```suggestion
    """Raised when manifests that must line up do not: WildASR family index
    alignment, or a fetched remote manifest that is not the full manifest with
    the requesting dataset's id.
    """
```

runner/src/coval_bench/datasets/manifests/README.md
```diff
@@ -346,6 +346,42 @@ print(json.dumps(manifest, indent=2, ensure_ascii=False))
 
 License: `Apache-2.0`.
 
+### `tts-v2.json` (private; pointer only)
```
callumreid, line 349: **P3 (ADVISORY) — docs/methodology.md's manifest enumeration no longer lists every manifest file.** methodology.md 'Versioning' says manifests live at `runner/src/coval_bench/datasets/manifests/{stt-v1,stt-v3,tts-v1}.json`; this PR adds `tts-v2.json` (a pointer) in that directory, so the enumeration is now factually incomplete even though tts-v2 runs nothing yet. Nothing in CI catches doc drift. Since BENCH-873 (dataset id as a setting) is the natural moment, fold a one-line update into it — add tts-v2.json to the list with a note that it is pointer-only until BENCH-873 turns it on.

callumreid (approved):
APPROVE — reviewed exact head d5a3a8021fa72ca343b1843e2a8b129a0e769b93 (PR is MERGED as of 2026-09-11T17:26:28Z, squash 8eb03e24, which is the current origin/main tip; CI evidence below is historical for this head). Scope: full three-dot diff from merge-base 723d7364 (7 files, +357/−87): GCS private-manifest pointer flow in datasets/loader.py + manifest.py, packaged tts-v2.json pointer, manifests/README.md catalog entry, the 2-line spoken-reference WER change in runner/orchestrator.py, and the loader/orchestrator test additions. Instruction sources: repo CONTRIBUTING.md (ruff, mypy --strict, pytest gates) and the skill checklist. The new-work delta since the approved review at fe94b1f7 (cc7dc7c9 + d5a3a802) is exactly the three requested polish items: the loader docstring correction, the items-or-remote exactly-one-of guard, and the guard tests — all verified in the diff and exercised by the new parametrized cases.

Validation: uv run pytest tests/unit/test_dataset_loader.py + the spoken-reference WER test + the Guava pin test + the Guava smoke test — 43 passed; uv run ruff check + format --check clean on all 5 changed files; uv run mypy --strict src tests — clean across 361 files. CI at the exact head: all 5 checks (Analyze Python, Runner CI, CodeQL, Plan checks, Switchboard Gate) success with head_sha verified = d5a3a802. Quality gates: WER reference selection is correct (compute_wer(reference, hypothesis) order verified against metrics/wer.py:150; stored wer_* counts follow the resolved reference; the result row keeps storing the hypothesis in `transcript` per existing convention); private fetches use storage.Client() ADC, never the anonymous client (loader.py:313 vs _make_storage_client, loader.py:370); credential handling logs only bucket/object names; content-addressing + pre/post-download sha256 verification + resolved-manifest id/alignment guards fail hard rather than mis-scoring; tts-v1 bytes and the orchestrator's tts-v1 hardcoding are untouched, matching the body's BENCH-873 note; wheel packaging confirmed (hatch packages src/coval_bench wholesale; the passing pointer test loads the packaged file).

PR body claims verified line-by-line against the diff; the only unverifiable-from-repo claim is the live-bucket upload (object exists at the content-addressed path with matching hash), which is an author-runtime statement — if wrong the loader raises DatasetIntegrityError at load, it cannot mis-score. Residual risk: Medium on live-GCS claims, Low on the code. Two P3 follow-up items inline (stale exception docstring; methodology.md manifest enumeration).

PR #632, opened 2026-09-11
[BENCH-883] Rebuild the shared pool after lifespan_pool closes it

coval-cale:
llm-sync reconciles the Coval agent inside one lifespan_pool and then runs the fetch inside another in the same process. psycopg refuses to reopen a closed pool, so the second entry raised PoolClosed and no LLM results were ingested. lifespan_pool now drops the process-wide reference when it closes the pool.

### DIFF
--- a/runner/src/coval_bench/db/conn.py
+++ b/runner/src/coval_bench/db/conn.py
@@ -43,7 +43,8 @@ async def get_pool(
 ) -> AsyncConnectionPool[psycopg.AsyncConnection[psycopg.rows.DictRow]]:
     """Return the process-singleton async connection pool.
 
-    The pool is created once per process (first call wins).  Callers are
+    The pool lives until ``lifespan_pool`` closes it, after which the next call
+    builds a fresh one; psycopg refuses to reopen a closed pool.  Callers are
     responsible for calling ``.open()`` before use and ``.close()`` at shutdown.
     Prefer ``lifespan_pool`` for FastAPI; use this directly in the runner
     orchestrator's startup/shutdown hooks.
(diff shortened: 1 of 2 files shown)


cooperr1 (approved):

callumreid (approved):
APPROVE — reviewed exact head 87b28ae32c1838e47e4997616e279f43986a5d7a, full three-dot diff from merge-base c214bc94 (2 files, +51/−1): the lifespan_pool finally-block that drops the process-wide _pool reference, plus a new regression test. Instruction sources: repo CONTRIBUTING.md (ruff, mypy --strict, pytest gates) and the skill checklist. No findings.

The failure mode in the body is real and I traced it end to end: sync-llm calls load_llm_models (coval_agent.py:145, lifespan_pool #1) which closes the pool but — before this PR — left _pool pointing at the closed instance; _run_fetch (coval_agent.py:334 → fetch_v2v.py:1301) then enters lifespan_pool #2, get_pool returns the closed pool, and pool.open() raises PoolClosed, so the ingest half of sync-llm died after the reconcile half — exactly 'no LLM results were ingested'. The fix is the minimal correct shape: after closing, null the global only if it still points at this pool (defensive against a replaced singleton). Every production pool close goes through lifespan_pool (verified all call sites: __main__.py ×6, api/app.py, orchestrator, s2s fetch, llm/coval_agent) so the cleanup covers the whole surface; psycopg close() is idempotent (verified from the installed psycopg_pool source) so an overlapping-lifespans double close is harmless, and no production path overlaps lifespans (each lifespan sits in its own asyncio.run). The new pool is created on the caller's event loop, so the cross-loop hazard does not arise.

Validation: uv run pytest tests/unit/test_db_conn.py — 1 passed (the regression pin: two sequential lifespans yield distinct, individually closable pools); ruff check + format clean on both changed files; uv run mypy --strict src tests — clean across 365 files (the test's type: ignore comments are used under CI's invocation; single-file mypy locally false-positives 'unused ignore' only because the package is not on that path — CI's real invocation is the one I ran). The full pytest suite is delegated to CI (pg_config not installed on this Mac); Runner CI ran it green at the exact head. Residual risk: Low.

PR #633, opened 2026-09-11
Store the S2S interruption rate unrounded

seribaymadina:
Each conversation's interruptions-per-minute value was rounded to two decimals in the fetch mapper before it reached the results table. Per-model means sit around 0.004 to 0.016 per minute, so the dashboard showed 0.00 or 0.02 and the bars looked empty. Coval's API already delivers the full float; this stores it as-is.

Rows already stored stay rounded. The bias that introduces on the mean is under 0.0002 per minute, so no backfill. The matching display change to three decimals is in benchmarks-web.

### DIFF
--- a/runner/src/coval_bench/s2s/fetch_v2v.py
+++ b/runner/src/coval_bench/s2s/fetch_v2v.py
@@ -520,9 +520,9 @@ def _expected_behavior_value(raw: object) -> tuple[float | None, ResultStatus] |
 
 
 def _interruption_value(raw: object) -> tuple[float | None, ResultStatus] | None:
-    """Interruptions per minute, stored as-is; a clip with no numeric value becomes a FAILED row."""
+    """Interruptions per minute, unrounded; a clip with no numeric value becomes a FAILED row."""
     if isinstance(raw, (int, float)) and not isinstance(raw, bool):
-        return round(float(raw), 2), ResultStatus.SUCCESS
+        return float(raw), ResultStatus.SUCCESS
     return None, ResultStatus.FAILED
 
 
--- a/runner/tests/unit/test_s2s_fetch.py
+++ b/runner/tests/unit/test_s2s_fetch.py
@@ -1649,7 +1649,11 @@ async def test_fetch_and_write_rejects_a_metric_with_no_row_builder(
 
 
 def test_interruption_value_maps() -> None:
-    assert fetch_v2v._interruption_value(1.1666666666666632) == (1.17, ResultStatus.SUCCESS)
+    assert fetch_v2v._interruption_value(1.1666666666666632) == (
+        1.1666666666666632,
+        ResultStatus.SUCCESS,
+    )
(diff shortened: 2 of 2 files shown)


coval-cale (approved):

runner/src/coval_bench/s2s/fetch_v2v.py
```diff
@@ -520,9 +520,9 @@ def _expected_behavior_value(raw: object) -> tuple[float | None, ResultStatus] |
 
 
 def _interruption_value(raw: object) -> tuple[float | None, ResultStatus] | None:
-    """Interruptions per minute, stored as-is; a clip with no numeric value becomes a FAILED row."""
+    """Interruptions per minute, unrounded; a clip with no numeric value becomes a FAILED row."""
     if isinstance(raw, (int, float)) and not isinstance(raw, bool):
-        return round(float(raw), 2), ResultStatus.SUCCESS
+        return float(raw), ResultStatus.SUCCESS
```
callumreid, line 525: **P3 (ADVISORY) — the display metadata that motivated this change still says 2 decimals.** METRIC_SPECS[INTERRUPTION_RATE].decimals is 2 in registries/metrics.py, and that registry's docstring says the metadata is consumed app-side against already-aggregated rows. With means at 0.004–0.016/min, any consumer honoring this registry still renders 0.00, which was the original bug the PR set out to fix. The body says the matching three-decimal display change lives in benchmarks-web — if benchmarks-web hardcodes its own precision there, this is moot; if it consumes this repo's METRIC_SPECS, the two repos now disagree and nothing in this repo's CI can catch it. Confirm benchmarks-web does not read this `decimals` value, or bump it to 3 here in the same change.

callumreid (approved):
APPROVE — reviewed exact head 4268234ece34eda995c2843ddbcf0b6978008ce3 (PR is now MERGED — it was merged at 2026-09-11T17:45:52Z by seribaymadina, after dispatch, with an approval from coval-cale at 17:29; head unchanged from the dispatched SHA, so this review stands and the finding below is post-merge follow-up material). Full three-dot diff from merge-base f355828a (2 files, +7/−3): _interruption_value stores float(raw) instead of round(float(raw), 2), plus test updates. Instruction sources: repo CONTRIBUTING.md (ruff, mypy --strict, pytest gates) and the skill checklist.

The change is correct and stores strictly more information: results.metric_value is DOUBLE PRECISION (init migration 0001), all aggregates are SQL AVG/percentile_cont (matviews 0005) with no rounding in the pipeline, the normalized dual-write reuses the same Result objects, and `decimals` from METRIC_SPECS is not consumed anywhere in this repo — so nothing in-repo assumes two-decimal stored values, and historical rows staying rounded (no backfill) is consistent and honestly disclosed. The updated test pins both the unrounded pass-through and a sub-0.01 value, and fails against the old code.

Validation: uv run pytest tests/unit/test_s2s_fetch.py — 100 passed; ruff check + format clean; mypy --strict clean on the changed module. CI: all 5 checks success at the exact head (head_sha verified per check-run). Residual risk: Low; one P3 advisory inline about display-metadata lockstep.

PR #95, opened 2026-06-11
Unify model registry

coval-cale:
## Summary by CodeRabbit

* **New Features**
  * Centralized model registry now drives provider/model availability across the app.

* **Bug Fixes**
  * Provider catalogue treats retired models as disabled and aligns listings with the registry.

* **Refactor**
  * Benchmark run configuration switched to registry-style model entries and matching by benchmark+provider+model.

* **Tests**
  * Test suite updated to validate registry invariants (uniqueness, voice rules, defaults).

<h3>Greptile Summary</h3>

This PR consolidates the previous two-file, dual-flag config (`ProviderEntry` with separate `enabled`/`disabled` booleans in `runner/config.py`) into a single unified `MODEL_REGISTRY` driven by a clean three-state `ModelStatus` enum (`ACTIVE`, `PAUSED`, `RETIRED`). All consumers — the orchestrator, the `/v1/providers` API, and the legacy migration importer — now pull from one authoritative source.

- **New `registries/models.py`**: defines `RegisteredModel` (frozen Pydantic model), `ModelStatus`, and the full `MODEL_REGISTRY` list; a `Counter`-based dedup check fires at import time to prevent silent duplicates.
- **Orchestrator**: override key now includes `benchmark` dimension, eliminating the previous ambiguity where a TTS deepgram entry could theoretically shadow an STT deepgram entry; `enabled_stt/tts` filtered to `ModelStatus.ACTIVE` only.
- **Tests**: `ProviderEntry` helper factories replaced with `_stt_entry`/`_tts_entry`/`_paused_registry`; new registry-invariant tests added (`test_registry_keys_unique`, `test_active_tts_models_have_voices`, `test_stt_models_have_no_voice`).

<h3>Confidence Score: 5/5</h3>

Safe to merge — the refactoring is internally consistent across all call sites, and the key behavioral change (ACTIVE-only filtering) is correct and well-tested.

The logic in all three consumers (orchestrator, API, migration) aligns correctly with the new registry semantics. The override key now correctly includes the benchmark dimension, fixing a latent cross-benchmark collision possible in the old two-key approach. Tests cover the uniqueness invariant, voice rules, and status-based filtering. No logic regressions were identified in the changed paths.

No files require special attention. The previous open threads around PAUSED model visibility in the API and import-source inconsistency remain latent but pre-date this PR's scope.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
[...]

PR #634, opened 2026-09-11
[BENCH-899] Save dashboard summaries and averaged timelines after runs

kdmelon:
## Summary
Dashboard reads currently recompute normalized summary statistics over individual observations. This saves exact 24h/7d/30d summaries at run completion and uses persisted hourly sufficient statistics for 7-day charts, 4-hour averages for 30 days, and hourly detail on drag-to-zoom.

The storage migration is isolated in [PR #645](https://github.com/coval-ai/benchmarks/pull/645). This application PR is based on that migration branch so its diff excludes the migration file. Completion queues, retry-safe source/hour replacement, backfills, and maintenance/repair commands keep the saved data current. Readers use one published generation, return explicit readiness errors, and expose freshness. Summary snapshots become stale after two hours, matching two hourly maintenance intervals. Existing metrics, summary percentiles, 24h run points, and the legacy read flag path are preserved. BENCH-884 remains separate.

If the runner image arrives before migration 0034, a savepoint isolates the missing-table enqueue error so the run's final status, timestamp, and error still commit. A structured warning identifies the skipped request; its source bucket requires explicit repair after migration. Other enqueue errors still propagate and roll back completion.

STT/TTS runs that finish failed skip completion-time summary publication; their durable source repair requests remain for hourly maintenance. Successful and partial runs still publish. Strict missing-hour readiness checks remain; infrastructure #197 adds failure and 90-minute no-success alerts.

## Review guide
Start with `dashboard_aggregates.py`, `dashboard_hourly.py`, and `dashboard_summaries.py` for source replacement and atomic publication. Then review the aggregate/leaderboard readers, completion savepoint, and rollout tests; the schema is isolated in #645.

## Testing
- Independent review of the completion guard, repair retention, alert configuration, and rollout instructions passed with no findings.
- Review fixes: 20 focused orchestrator, PostgreSQL maintenance, and API snapshot tests pass; the new failed-run assertion reproduced the issue before the fix. Ruff lint/format and strict mypy pass on all three changed Python files.
- Standalone runtime smoke: the real orchestrator and database writer persisted failed/succeeded/partial/failed runs with local provider stubs; summary generations advanced by 0/1/1/0. The disposable PostgreSQL server stopped afterward.
- Final stacked branch: 129 leaderboard/snapshot/S2S tests pass; the standalone migration/completion/repair smoke and merged-file lint, format, and strict typing pass. Independent review found no merge issues.
- Completion compatibility: all three status cases reproduced the missing-table failure before the fix; 9 maintenance tests and 51 existing legacy/normalized writer tests pass after it. Coverage includes actual schema 0033 and rollback on a real queue constraint violation.
- Standalone PostgreSQL smoke passes: pre-migration completion, independent migration-only branch upgrade, post-migration enqueue, real repair CLI recovery of a skipped bucket, retained observations, and compatibility with the unchanged writer. PostgreSQL stopped afterward.
- Changed-file Ruff lint/format and strict mypy pass. Independent transaction-safety review completed before publication.
- Previously verified: 16 hourly freshness tests, affected API/backfill/hourly/summary/completion tests, and Alembic → RunWriter → maintenance CLI → four HTTP endpoint smoke.

## Rollout
Merge and apply migration PR #645 first. Retarget this PR to `main` after #645 merges, then merge and deploy the application with normalized reads disabled. Repair requests skipped during any deployment-before-migration gap, initialize saved data, and verify recurring refreshes and production latency before cutover. Deploy frontend #95 alongside or immediately after this PR because legacy 7d/30d timelines also switch to averages before normalized-read cutover. The companion infrastructure PR #197 adds a paused hourly maintenance job plus failure and 90-minute no-success alerts; finish initialization and resume scheduling in the same rollout because the heartbeat also alerts before the first success. See `docs/write-time-dashboard-aggregates.md`.

No production migration, deployment, or flag change has been performed by this task. Tracks BENCH-899 under BENCH-882.

Metric identity changes are isolated in [PR #649](https://github.com/coval-ai/benchmarks/pull/649). This PR has been restored to its original aggregate-only source tree.


coval-cale (approved):
Reviewed at c20cc34a. No actionable findings. All 466 relevant tests passed, including PostgreSQL, API, writer, backfill, and completion-hook tests. Production refresh load and query latency remain rollout validation items.

runner/src/coval_bench/runner/orchestrator.py
```diff
@@ -1552,6 +1552,12 @@ def _on_sigterm() -> None:
             if final_status in (RunStatus.SUCCEEDED, RunStatus.PARTIAL):
                 await _refresh_series_bucket(writer, run_id, settings)
 
+            try:
+                snapshot_status = await writer.refresh_dashboard_summaries(run_id)
```
ben-coval: This runs for every terminal status, including FAILED, where nothing new landed in the source tables. Each call is three REFRESH MATERIALIZED VIEW CONCURRENTLY passes over 30 days of observations with a 550s statement timeout, awaited before the run summary is written. Suggest moving it inside the `if final_status in (SUCCEEDED, PARTIAL)` block above, or dropping the completion hook entirely and letting the hourly job own publication (the `stale` flag already covers the lag).

runner/src/coval_bench/api/routers/aggregates.py
```diff
@@ -339,6 +352,193 @@ def _mean_split(column: str) -> str:
...
+            AND s.definition_fingerprint=%(fingerprint)s""",
+            {
+                "start": full_start,
+                "end": full_end,
+                "revision": DEFINITION_REVISION,
+                "fingerprint": aggregation_fingerprint(),
+            },
+        )
+    ).fetchone()
+    if row is None or row["present"] != row["expected"]:
```
ben-coval, line 510: Once the flag is on, this makes 7d/30d normalized timelines a hard dependency on the maintenance job. Hours with no run completions only get a `dashboard_hourly_state` row from reconcile, so if the (born-paused) scheduler lapses for more than an hour, every full hour after that is absent and this 503s for everyone.

A missing row is distinguishable from a dirty one: run completion always inserts a dirty row for its hour, so absent means no source ever touched it. Treating absent as an empty hour and setting `stale: true` would degrade instead of fail. If you would rather keep the strict check, #197 should ship with an alert on job failure or no successful execution in ~90 minutes.

ben-coval (approved):
Reviewed at a3a2256, including the metric-ID delta and the finish_run savepoint. Two things below I would like addressed before or shortly after merge; neither blocks. Deploy benchmarks-web #95 right after this, since legacy 7d/30d timelines switch to averaged buckets regardless of the normalized flag.

PR #638, opened 2026-09-11
ci: avoid duplicate ready-for-review runs

borgesius:
## Summary

Stop redispatching the Switchboard gate when a draft PR is marked ready.

## Why

Draft opened events and every synchronize event already run this workflow. Marking the PR ready does not change its head SHA, so the extra dispatch duplicates exact-head CI without adding coverage.

## Validation

- actionlint on the changed workflow

### DIFF
--- a/.github/workflows/switchboard.yml
+++ b/.github/workflows/switchboard.yml
@@ -3,7 +3,7 @@ name: Switchboard
 on:
   pull_request:
     branches: [main]
-    types: [opened, synchronize, reopened, ready_for_review]
+    types: [opened, synchronize, reopened]
   merge_group:
     types: [checks_requested]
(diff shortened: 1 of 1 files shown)


callumreid (approved):
OINK — approved; this is the same one-line fix as the already-approved backend#7358, verified against this repo's own trough: drafts already run these checks (draft #634's head has completed Plan-check runs, so marking ready was a same-head re-dispatch), the `switchboard-pull_request-<head_ref>` concurrency with cancel-in-progress just restarts identical work, and no other workflow in the repo keeps `ready_for_review`. Merge-group and workflow-dispatch triggers untouched; counts match `gh pr view` (1 file, +1/−1).

PR #608, opened 2026-09-09
feat(tts): add Airy streaming provider

fbdp1202:
## Summary

- Add support for `airy-tts-v1`, the TTS model offered through Out of Set's Airy service.
- Stream audio over HTTP from `https://api.airy.so/v1/audio/speech/stream` using 24 kHz mono signed 16-bit little-endian PCM.
- Use shared HTTP client and TTS finalization helpers for WAV output, time-to-first-audio measurement, and connection diagnostics.
- Register the provider and document the optional `AIRY_API_KEY` environment variable.
- Validate response audio metadata and reject failed, interrupted, or incompatible streams.

Closes #609

## Files changed

| File | Purpose |
| --- | --- |
| `runner/src/coval_bench/providers/tts/airy.py` | Airy streaming TTS implementation |
| `runner/src/coval_bench/providers/tts/__init__.py` | Provider registry and exports |
| `runner/src/coval_bench/registries/provider_keys.py` | API key environment mapping |
| `runner/src/coval_bench/config.py` | Optional Airy API key setting |
| `.env.example` | Document `AIRY_API_KEY` |
| `runner/tests/providers/tts/test_airy.py` | Request, audio output, timing, authentication, and error-handling tests |

## Suggested benchmark voices

Please use the following voices when configuring Airy benchmark runs:

| Voice | Voice ID |
| --- | --- |
| Female | `a597bb7a98fc9ec1` |
| Male | `bdb7de5e2cdd3324` |

## Test plan

- [x] Airy provider unit tests: 17 passed.
- [x] Ruff lint and format checks passed.
- [x] `mypy --strict src tests` passed.
- [x] Full pytest suite: 2,047 passed, 6 skipped due to unavailable optional Google SDKs and separate test fixtures.
- [x] Live API smoke test: HTTP 200 over HTTP/2 with valid 24 kHz mono 16-bit WAV output.

PR #639, opened 2026-09-14
fix(tts): use Airy voice payload key

fbdp1202:
Follow-up to #608.
Updated the Airy TTS request payload to use the `voice` key instead of `voice_id`, matching the latest API contract.
A dedicated API key was shared separately, and I also reached out via email at yechan@outofset.com.


callumreid (approved):
READY — APPROVE.

What was reviewed: base origin/main (freshly fetched) vs head c2cb746, one commit, two files (runner airy.py payload key, test_airy.py assertion). Instruction sources: benchmarks CONTRIBUTING.md (ruff, mypy --strict, pytest required), runner pyproject/ruff config, existing Airy provider code from #608.

The `voice_id` → `voice` rename is verified against the live vendor contract: Airy's published docs for POST /v1/audio/speech/stream list request fields input/language/model/voice/style, and the docs' own cURL example matches the new payload exactly (including `"voice": "a597bb7a98fc9ec1"`, which is the vendor's documented default voice id — the test's _VOICE fixture, public data, not a credential). Response parsing (audio/pcm + X-Audio-* headers) matches the docs and is untouched. No stale `voice_id` remains in any Airy context; other providers' `voice_id` payloads are their own contracts. test_airy.py asserts exact payload dict equality, so it fails against the old key.

PR body accuracy: "matching the latest API contract" — confirmed true against today's docs. "A dedicated API key was shared separately" — consistent with the GCP secret `benchmarks-airy-api-key` (env AIRY_API_KEY) declared in benchmark-infra; no key value appears anywhere in the diff, tests, or body. The email sentence is the author's public contact info — harmless.

Validation evidence: exact-head CI green — check-runs API confirms all 5 checks (Analyze Python, Runner CI, CodeQL, Plan checks, Switchboard Gate) completed successfully against c2cb746; Runner CI includes ruff check, ruff format --check, mypy --strict, full pytest with coverage, pip-licenses, gitleaks, and both docker builds. Locally: `uv sync --frozen --dev` + `uv run --frozen --dev pytest tests/providers/tts/test_airy.py` → 17 passed.

Residual risk: Low — payload verified against the vendor's published docs rather than a live synthesis call (a live call would spend credits and use a production key); exact-head CI and the runtime pipeline cover any residual drift.

fbdp1202:
Hi Coval team, thanks again for reviewing this PR!
A quick update: Airy TTS API is now live, and I wanted to share a few highlights:
- Time to first audio (TTFA): Under 100 ms in our internal measurements.
- Generation speed: Approximately 2,300 characters per second in our internal US-West-2 test: a 495-character English input produced 27 seconds of audio, with the complete response received in 215 ms.
- Pricing: $5, ~$3.67, and $2 per million English characters on Plans A/B/C ($11/$99/$480 per month), respectively, at full use of included credits.
These are internal results, and we’d really value seeing how Airy performs under Coval’s benchmark methodology.
When you have a chance, could you merge this fix and run the benchmark with Airy? I’ve shared a dedicated API key separately and would be happy to resend it privately or help with setup.
[Docs and pricing](https://airy.so/cloud-api/docs/home) · [Voice samples](https://airy.so/cloud-api/docs/guide/voices)
Thanks for your time and for maintaining these benchmarks!

PR #640, opened 2026-09-14
[BENCH-897] Serialize requests per model and run models in parallel

coval-cale:
Replace the single semaphore with a ModelGate: one lock per (provider, model) under the existing global cap of 8. A model never serves two of our requests at once, and different models proceed side by side. The dedicated cap of 1 goes away since the per-model lock is what it protected; the nightly Baseten suite streams to both STT endpoints concurrently instead of alternating.


runner/src/coval_bench/runner/gate.py
```diff
@@ -0,0 +1,39 @@
+"""Per-model concurrency gate for provider requests.
```
seribaymadina, line 1: wonder if we should add metodology marker note, if our hypothesis about this affecting ome of the latency for certain providers due to the queue?

seribaymadina:
**[P2] Release the TTS model lock before scoring and uploads.** The lock at [`orchestrator.py:837`](https://github.com/coval-ai/benchmarks/blob/0ff90a8e8ad1ffa677c151258133384b1d7016cd/runner/src/coval_bench/runner/orchestrator.py#L837) stays held through Whisper scoring, database writes, and artifact uploads. A slow scoring request therefore blocks the next synthesis for that model even though the TTS endpoint is idle, unnecessarily extending the run. Could we release the model lock after synthesis and keep post-processing bounded separately by the global cap?

seribaymadina (approved):
very cool! left couple comments

runner/src/coval_bench/runner/gate.py
coval-cale: Shouldn't affect latency, but I'll check back in tomorrow and add a methodology marker if it does.

PR #189, opened 2026-06-25
Add a provider select-all checkbox to the model sidebar

coval-cale:
Toggles every model under a provider at once, with an indeterminate state when only some are selected.

Just a hot fix until proper filtering lands.

## Summary by CodeRabbit

* **New Features**
  * Added provider-level bulk selection in the model sidebar, letting you select or clear all models for a provider at once.
  * Provider checkboxes now reflect selection state with checked and indeterminate visuals for clearer feedback.

* **Bug Fixes**
  * Improved selection behavior so provider toggles stay in sync with the currently selected models.
  * Updated accessibility labels to better describe the provider selection state.

<h3>Greptile Summary</h3>

This PR adds a select-all checkbox to each provider group in the model sidebar. Clicking the header checkbox selects or deselects all models under that provider at once, with a visually correct indeterminate state when only some of the provider's models are selected.

- **`useDashboardState`**: Adds `toggleProviderSelection`, which computes the next selection by either union-ing or filtering, fires the existing PostHog analytics event, and calls `setSelectedModels` — consistent with the `toggleModelSelection` pattern already in place.
- **`ModelSidebar`**: Converts the plain provider heading into a `<label>` wrapping a custom styled checkbox, uses a ref callback to imperatively set `element.indeterminate`, and renders a checkmark or dash SVG via Tailwind's `peer-checked` / `peer-indeterminate` variants to match the visuals of the existing per-model checkboxes.

<h3>Confidence Score: 4/5</h3>

Safe to merge; the new provider-level checkbox works correctly and the hook logic is well-guarded against empty provider lists and duplicate entries.

The select/deselect logic is correct, the indeterminate state is properly driven through the ref callback, and the Set-based deduplication in the select path prevents duplicates. The only notable pattern is the inline ref function being recreated on every render, which triggers React's cleanup+setup cycle per provider row per selection change — not a bug, but worth tidying up when the provider list grows.

web/components/layout/ModelSidebar.tsx — the inline ref callback for `indeterminate` is the one spot that could benefit from a stable reference if re-render frequency becomes noticeable.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| web/hooks/useDashboardState.tsx | Adds toggleProviderSelection callback; logic is correct, deselect path uses filter + Array.includes (O(n·m) but negligible at this scale), PostHog event fires before setSelectedModels, consistent with toggleModelSelection. |
[...]

PR #643, opened 2026-09-14
Drop the S2S industry boards from the runner

seribaymadina:
The board moved to the Ultra Bank set on Sept 11 and infra #189 already stopped fetching the Health, Home Service and Customer Service runs, but the runner still carried their three families, nine agent specs, sixteen settings and a separate Expected Behavior Adherence fetch path. This removes all of it. The bank board stores the expected-behaviors fraction directly as instruction adherence, so the second metric had nothing left to add.

Dental and bank families, stored rows, the samples partition and the `EXPECTED_BEHAVIOR_ADHERENCE` enum are untouched. A follow-up infra PR drops the matching env block.


coval-cale (approved):
Love a big deletion

PR #646, opened 2026-09-14
[BENCH-902] Share the bank scenario between S2S and LLM

coval-cale:
## What

One shared scenario definition for both benchmarks, and persona ids out of the public repo.

- `scenarios.py` replaces the LLM-only `llm/suite.py`. It names the bank contract, the dataset family for each benchmark, and the three settings holding the Coval test set, judge, and clean persona ids.
- Both spec builders read it. The four S2S bank agents and the LLM specs no longer repeat setting names per agent.
- The leaderboard derives each benchmark's primary dataset from the scenario instead of a literal map.
- Only text runs require the persona setting. Voice runs select personas through the condition map, so `missing()` takes the benchmark.
- The samples card's persona-to-label map is gone from `s2s/samples.py`. Labels come from a new `s2s_persona_labels` setting threaded into the publisher. Tests use placeholder ids. No persona id remains in source or tests.

## Why

LLM and S2S are the same evaluation with different inputs and outputs, so the scenario should be named once. Personas are part of the evaluator, not the agent, so their ids belong with the other private ids in benchmark-infra.

## Reviewer notes

- Depends on the infra change on BENCH-901 landing first: it adds `COVAL_S2S_BANK_PERSONA_ID` to both jobs and `S2S_PERSONA_LABELS` to s2s-fetch. Until then llm-sync fails at startup naming the missing setting, and sample manifests fall back to raw persona ids as labels.
- No behavior change for S2S ingestion beyond the label source.
- Full suite with the CI socket flags: 2258 passed. ruff and mypy strict clean.

Linear: BENCH-902


seribaymadina (approved):
amazing love thiss

PR #655, opened 2026-09-15
[BENCH-915] Fix ignored LLM run-template updates

coval-cale:
## Summary

LLM run-template updates used singular ID fields that Coval silently ignored, leaving dental callers paired with the Ultra Bank agent. Send the supported plural arrays and verify the returned template matches all managed fields, failing sync if an update was ignored or only partially applied.

The existing sync tests now cover successful reconciliation, idempotency, and ignored updates.

## Validation

- 20 focused tests passed; Ruff and strict mypy passed.
- All five PR checks pass, including Runner CI.

The public API reference still documents the old singular fields; correcting the published schema is a separate cross-repository follow-up.

Fixes BENCH-915.


runner/src/coval_bench/llm/benchmark.py
```diff
@@ -104,12 +99,4 @@ def run_template_body(
...
-    body: dict[str, Any] = {}
-    for path in paths:
-        singular = _TEMPLATE_PATCH_KEYS.get(path)
-        if singular is None:
-            body[path] = wanted[path]
-            continue
-        (only,) = wanted[path]
-        body[singular] = only
-    return body
+    return {path: wanted[path] for path in paths}
```
callumreid, line 102: **P3 (advisory, cross-repo) — the public API reference still documents the singular schema that caused this bug.** The `docs` repo's `api-reference/v1/run-templates-v1.yaml` (auto-synced from the published OpenAPI spec) documents `agent_id`/`persona_id`/`test_set_id` on the Create/Update requests and the resource, but the live v1 API only accepts and returns the plural arrays (`agent_ids`/`persona_ids`/`test_set_ids`) — `persona_ids` appears nowhere in the docs repo on main. The singular mapping removed here (BENCH-825) was evidently written against that stale spec, and the permissive mock let it survive. Worth a docs-side PR (or a fix to whatever generates the published spec) so the next integration doesn't repeat this. Not blocking; no change needed in this PR.

callumreid (approved):
**APPROVE** — the fix is correct, closes the whole ignored-update path (not a happy-path subset), and adds a read-after-write reconciliation check with a regression test that pins both halves.

**What was reviewed:** single-commit PR at head 8d91659960933a00547979bd3dca838aec916ad1; merge-base 8d728718 is the current main tip, so HEAD..origin/main is empty (no staleness). Scope: runner/src/coval_bench/llm/benchmark.py, runner/src/coval_bench/llm/coval_agent.py, runner/tests/unit/test_coval_agent_sync.py. Instruction sources: CONTRIBUTING.md (ruff + mypy --strict + pytest gate), runner/pyproject.toml, neighboring code.

**Root cause confirmed against the live API:** the old template_patch_body sent singular scalar keys (agent_id/persona_id/test_set_id), but the v1 UpdateRunTemplateRequest (backend src/modules/domain/run_templates/common/schemas.py on origin/main) accepts only the plural array fields, and pydantic silently drops unknown keys — so every run-template PATCH was a 200 no-op while sync reported a patch. The permissive test mock merged the singular keys as new top-level fields, which is why the old test masked it. The fix sends exactly the plural keys the create path has always used (and the create schema requires), and the post-patch re-plan over all TEMPLATE_MANAGED fields turns any future silently-ignored or partially-applied update into a loud SyncError. The backend maps the PATCH response from the persisted row inside the same transaction, so the verification reads true persisted state, not an echo. Backward compatibility is safe: templates have only ever been created with plural keys, and legacy single-config templates are exposed as one-element arrays matching the wanted single-element lists — the first sync after deploy simply applies the drift that silently accumulated.

**Validation:** all five checks green against this exact head_sha (Runner CI = ruff check + ruff format + mypy --strict + pytest with coverage + license/secret scans + docker build, Switchboard Gate, CodeQL, Analyze Python, Plan checks). Locally, read-only and without installs: python3 -m py_compile on the three changed files OK; ruff check . and ruff format --check . from runner/ (invoked exactly as CI does) both clean. pytest/mypy could not run locally — system Python is 3.14 without the pinned >=3.12,<3.13 dependency set and installs are out of scope for this review — so those rely on the exact-head CI run. The regression test was traced by hand: it fails without the key fix (verification raises in the apply branch) and fails without the verification (nothing raises in the server-ignores branch).

**Quality-gate summary:** correctness traced across dry-run, apply, create, verify, error, and retry paths (PATCH is idempotent so the daily cron retries safely); no new swallowed errors — SyncError surfaces through sync_llm's existing RUN_FAILED/ClickException handling; no migration; conventions, comments, and secret hygiene clean; tests assert behavior rather than implementation trivia. Residual risk is low: the verification trusts the PATCH response rather than a re-fetch GET (sufficient for the current in-transaction backend mapping), and one provider's failed reconciliation now fails the whole invocation closed, skipping that day's ingestion — the right trade for benchmark integrity versus silently collecting runs from a misconfigured template. Minor: the PR body is empty while sibling PRs in this repo carry descriptive bodies; the title and commit subject are accurate.

PR #656, opened 2026-09-15
[BENCH-916] Populate normalized metric IDs and add resumable backfill

kdmelon:
## Summary

Populate shared catalog metric IDs across normalized application writes, aggregation, reads, and explicit historical backfills.

- Resolve and validate IDs in writers, preserve retry identity, and carry IDs through projections and source/hourly aggregation. Concurrent historical retries fill only missing IDs and read back the winning update.
- Preserve public metric codes and temporarily support historical NULL IDs in normalized reads.
- Add a default-read-only backfill command with explicit apply mode, bounded committed batches, ownership locking, interruption/resume, and reconciliation. Extend readiness checks.

Fixes BENCH-916

## Dependency

The schema migration is isolated in [PR #657](https://github.com/coval-ai/benchmarks/pull/657) (BENCH-919). PR #657 is merged and this PR now targets `main`, with the merged migration incorporated. Apply revision `20260915_0036` before deploying this application change.

## Review guide

1. Review `db/writer.py`, `db/dashboard_source.py`, `db/dashboard_hourly.py`, and `api/routers/aggregates.py` under `runner/src/coval_bench/` for code/ID agreement and historical fallback. Existing UUID links, code conflict keys, metric dimensions, public codes, and summary definitions remain intact.
2. Review `runner/src/coval_bench/migrations/backfill_normalized_metric_ids.py` for transaction boundaries, ownership, failure handling, and final verification.

## Testing

- Fixed all three CI failures by adding the metric-ID coverage query to the readiness test connection. Readiness module: 21 passed; changed-file strict mypy, Ruff, and format checks passed. Remote CI is rerunning for this head.
- Deterministic PostgreSQL concurrency regression reproduced the old writer failure in all four lifecycle states; all four cases pass with the repair and preserve every non-ID field.
- Full normalized-ID module: 19 passed. Focused writer/backfill checks: 31 passed.
- Full CI type-check command (`mypy --strict src tests`): all 398 source files passed after fixing six test-typing errors. Changed-file Ruff/format and diff checks passed.
- Standalone runtime smoke passed: historical migration/backfill → fresh writer → source/hourly/summary publication → aggregate HTTP 200. Disposable PostgreSQL was stopped.
- Existing unaffected aggregation/API/leaderboard checks and prior data-integrity review are reused. Independent repair review: passed with no blocking findings.

## Rollout

Historical backfill is an explicit operator step, separate from migration and startup. BENCH-917 enforces NOT NULL and ID retry keys after verified backfill coverage; BENCH-918 removes compatibility later. Legacy tables are excluded. Production migration is operator-managed; this PR does not run migration or backfill automatically.


runner/src/coval_bench/db/writer.py
```diff
@@ -439,6 +475,16 @@ async def insert_metric_evaluation(
                         ),
                     )
                     row = await cur.fetchone()
+                    if row is not None and row["metric_id"] is None:
+                        await cur.execute(
+                            """UPDATE benchmarks_v2.metric_evaluations
+                               SET metric_id = %s WHERE id = %s
```
ben-coval: This UPDATE has no `metric_id IS NULL` guard, so two concurrent retries of the same un-hydrated historical row can both reach it. The second one sets `metric_id` to the value it already has, which means `OLD.metric_id` is no longer NULL, the 0036 hydration exception doesn't match, and `validate_metric_transition` fires and raises (`invalid work status transition from queued to queued`, or `terminal work rows are immutable` for succeeded/failed rows).

Narrow window and it self-heals on the next retry, but the fix is small:

```sql
UPDATE benchmarks_v2.metric_evaluations
   SET metric_id = %s
 WHERE id = %s AND metric_id IS NULL
RETURNING ...
```

and if that returns no row, re-run the SELECT above so `row` isn't left as `None` for a row that does exist.

ben-coval (approved):

PR #111, opened 2026-06-12
Default STT metric to TTFS and sync the toggle across all latency charts

seribaymadina:
Makes TTFS the default metric on the STT dashboard and ties the TTFS/TTFT switch to one shared state so the whole page moves together.

Toggling on any chart now updates the summary headline, timeline, box plot (Latency Variation), latency vs accuracy scatter, and heatmap at the same time. The box plot and heatmap previously hardcoded TTFT and now follow the toggle. The WER bar chart is left alone since it is not a latency metric. TTS is unaffected and stays on TTFA with no toggle.

No backend or data changes. All TTFS values were already present in the stats payload.

## Release Notes

* **New Features**
  * Introduced unified metric toggle control for STT metrics (TTFS/TTFT) across dashboard sections
  * Charts and visualizations now dynamically update based on the selected metric

* **UI/UX Improvements**
  * Consolidated metric selection into a consistent toggle control across box plots, heatmaps, latency/accuracy, and timeline charts
  * Metric selection synchronized dashboard-wide for improved user experience

<h3>Greptile Summary</h3>

This PR makes TTFS the default STT metric and unifies the TTFS/TTFT toggle into a single shared piece of state (`sttMetric` / `activeMetric`) so that every latency surface — the summary headline, timeline, box plot, scatter plot, and heatmap — moves together when the toggle is clicked on any chart.

- **`useDashboardState`**: adds `sttMetric` state (default `\"TTFS\"`) and derives `activeMetric` from it; threads `activeMetric` through headline, heatmap memoisation, and box plot calls.
- **`useChartData`**: re-indexes box-plot and heatmap data by metric type up front (`boxByMetricModel`), and converts `getBoxPlotData`/`getHeatmapData` from memoised values to `useCallback`s that accept a metric argument, enabling on-demand lookup without extra passes.
- **`MetricToggle`**: new shared component that reads/writes `sttMetric` from the dashboard context and renders `null` on TTS, replacing four copies of the inline toggle buttons.

<h3>Confidence Score: 5/5</h3>

Safe to merge — the change is purely frontend state consolidation with no backend or data-schema impact.

All five latency surfaces now read from a single reactive activeMetric value derived from sttMetric. The memoisation is applied correctly at each consumer, the getBoxPlotData/getHeatmapData callbacks have accurate dependency arrays, TTS remains unaffected because activeMetric is hardcoded to TTFA when page === tts, and MetricToggle guards against appearing on the TTS tab.

No files require special attention.

[...]

PR #659, opened 2026-09-15
[BENCH-921] Add results API v2

kdmelon:
## Summary

Add `GET /v2/results` backed by metric evaluations, with stored primary values, optional named components, independent run/evaluation status filters and cursor pagination. Catalog lookup uses `metric_id` first and falls back for historical null IDs, so the metric-ID backfill is not a prerequisite for reads. `/v1/results` remains available with its existing contract.

Fixes BENCH-921
Companion client types: [benchmarks-web #111](https://github.com/coval-ai/benchmarks-web/pull/111). Regenerate them against the deployed v2 OpenAPI before adopting the endpoint; frontend builds generate types from the running API.

## Review guide

1. Start with `runner/src/coval_bench/api/routers/results_v2.py`: visibility and query filters select the evaluation page before values/components load, and metric lookup supports both populated and historical null IDs.
2. Review the response schemas and focused results v2 tests for primary-role selection, nullable values, version/status identities, pagination, visibility and mixed pre-/post-migration data.
3. Read `docs/raw-results-api.md` for public usage examples, response defaults, cursor/late-arrival semantics, and v1 migration behavior.

## Testing

- Focused API tests: **67 passed**, including all 16 v2 cases plus v1 results, access/visibility and PostHog tests. Metric-ID migration tests: **19 passed**. These behavior checks are reused for the identifier-only rename.
- Disposable PostgreSQL HTTP smoke passed with pre-0036 history and post-0036 fresh evaluations, pagination, components and bounded reads. Historical IDs remained null after reads. Actual endpoint SQL was checked with `EXPLAIN ANALYZE BUFFERS` in both component modes; database cleanup was confirmed.
- Changed-file Ruff lint/format and Python compilation passed after the rename.
- Fresh HTTP `/openapi.json` smoke passed. Comparing the complete schema before/after found only the intended operation/model name changes; parameters and response fields are identical, operation IDs are unique, and references resolve. Null primary values and omitted/empty components retain their serialization behavior.
- Regenerated the companion TypeScript definitions and passed the focused strict typecheck.
- Independent review passed for the implementation, metric-ID integration and naming update. Public docs were inspected for client-facing usage and kept to 69 lines.

## Known limitations and rollout

- Apply migration `0036` before deploying v2. Metric-ID backfill may follow; API reads do not perform it.
- Verify production source-table read permissions, evaluation history coverage and representative-volume latency before release. Local query-plan evidence uses a small disposable database.
- Full local mypy validation is limited by existing OpenTelemetry/package import typing issues; no dependency changes are included.


coval-cale (approved):

ben-coval (approved):
Read the router, schemas, tests, and docs against BENCH-921. Contract matches the ticket and the cursor/visibility handling looks right.

One non-blocking note: the catalog join goes through `metric_id_for_code(e.metric_type)`, which raises on an unknown code. If any pre-0036 evaluation row carries a code that is not in the catalog, every v2 page that scans it will 500 rather than skip it. Writers only produce catalog codes today, so this is unlikely, but a plain `LEFT JOIN metrics ON code = metric_type` with `COALESCE(e.metric_id, m_code.id)` would avoid the failure mode and let the planner use the unique index on `code`.

kdmelon:
@ben-coval am finishing backfill tonight so it shouldn't be an issue

PR #660, opened 2026-09-16
[BENCH-925] Add GPT-5 minimal and medium benchmarks

coval-cale:
Support separate GPT-5 minimal and medium Ultra Bank benchmarks with registry-owned configuration. Each LLM registry entry supplies its upstream model and reasoning effort; runtime routing contains no model-name or reasoning alias maps. Model-specific clients, Coval identities, and ingestion keep results separate while preserving existing agents.

Migration 0037 adds the configuration field and backfills existing GPT-4.1, Gemini, and Phonely settings. It does not register new models. After deploying the compatible API and llm-sync runner and applying the migration, register both GPT-5 variants through the admin API, run llm-sync, launch fresh benchmarks, and sync the completed results.

Validation: latest targeted registry, admin API, and migration rollback tests: 52 passed. Ruff and strict mypy pass. Tests cover registration of both GPT-5 variants through the admin API, registry configuration round-trips, routing after configuration edits, preserved agent identities, and per-model TTFT attribution. Earlier full suite had 2,406 passing and four migration failures; all four passed after rollback fixes. Both reasoning settings previously passed live streaming smoke calls.

https://linear.app/coval/issue/BENCH-925/benchmark-gpt-5-with-minimal-and-medium-reasoning


runner/src/coval_bench/api/routers/admin_models.py
```diff
@@ -206,6 +210,11 @@ async def update_admin_model(
         raise HTTPException(404, "no such model")
     candidate = ModelRecord.model_validate({**current.model_dump(), **changes})
     await _validate_model(candidate, store)
+    if current.llm_config is not None and (
+        candidate.llm_config is None
+        or candidate.llm_config.legacy_provider_route != current.llm_config.legacy_provider_route
+    ):
+        raise HTTPException(422, "keep the existing LLM agent identity; pause with collected=false")
```
callumreid, line 217: **P2 — a rename can still fork the LLM agent identity this guard exists to protect.** The guard blocks `llm_config=None` and the `legacy_provider_route` flip, but for non-legacy rows `CovalTextAgentDefinition.identity` is `sha256(model)[:16]` (`llm/coval_agent.py:63-64`), so a PATCH renaming `model` or `provider` on a row that has `llm_config` silently changes `customer_agent_id`. The next `sync-llm` then creates a brand-new Coval agent and scheduled run, while the old agent keeps its enabled daily schedule and every run 404s at the proxy because the roster no longer matches — nothing reconciles or disables it, and `_rename_warnings` covers METRIC_EXCLUSIONS/retired keys/pricing but not LLM identity, so the rename goes through with no warning at all. Existing validation misses it because the check compares only the config, not the identity-bearing fields. Smallest sound fix — include them in the same guard:

```suggestion
    if current.llm_config is not None and (
        candidate.llm_config is None
        or candidate.provider != current.provider
        or candidate.model != current.model
        or candidate.llm_config.legacy_provider_route != current.llm_config.legacy_provider_route
    ):
        raise HTTPException(422, "keep the existing LLM agent identity; pause with collected=false")
```

If renames must stay possible for LLM rows, instead add an LLM identity entry to `_rename_warnings` naming the agent/schedule the rename orphans.

runner/src/coval_bench/api/routers/admin_models.py
```diff
@@ -73,6 +73,10 @@ def _implemented_providers(modality: Benchmark) -> frozenset[str] | None:
 
 async def _validate_model(candidate: NewModel | ModelRecord, store: RegistryStore) -> None:
     """The write-time rules the registry's review used to enforce; 422 on violation."""
+    if candidate.llm_config is not None and candidate.modality is not Benchmark.LLM:
+        raise HTTPException(422, "llm_config is LLM-only")
+    if candidate.modality is Benchmark.LLM and candidate.collected and candidate.llm_config is None:
+        raise HTTPException(422, "a collected LLM model needs llm_config")
```
callumreid, line 79: **P3 — a duplicate legacy route surfaces as a misleading 409.** `_validate_model` accepts a row with `llm_config.legacy_provider_route=true` even when another LLM row for the same provider already claims the provider-only route; the partial unique index `models_llm_legacy_provider_route` from 0037 then rejects the write, and `insert_model`/`update_model` (`db/registry_store.py:307-310, 399-402`) translate every UniqueViolation to "model (LLM, openai, X) already exists" — an admin gets a 409 naming a model that does not exist. Tests can't catch this because the admin-API test schema (`tests/api/conftest.py`) omits the partial index, so the 409 path is never exercised. Either reject it here with a precise 422 (scan `await store.list_models()` for a same-provider LLM row with a legacy route, excluding `candidate.id` on PATCH), or map the constraint name to an honest message in the store's DuplicateKey translation.

callumreid (approved):
**Verdict: APPROVE.** One non-blocking P2 (guard-completeness gap on an exceptional admin action) and one P3 ride inline; neither has a failure scenario in the flows this PR itself performs, so neither meets the blocking bar.

**Reviewed:** base = current `origin/main`, head `a75943d` (worktree SHA verified matching, not stale), full 4-commit diff (+602/−100, 18 files — file list cross-checked against GitHub, exact match). Conventions derived from existing code (migrations 0020/0021/0025/0029/0036, `registry_store`, admin router, existing tests) — the repo has no AGENTS.md.

**Validation evidence:** exact-head CI 5/5 green (Runner CI runs ruff, ruff format, strict mypy, the full pytest suite with embedded Postgres, and both Docker builds; plus CodeQL, Switchboard Gate, Plan checks, Analyze Python). Locally: `py_compile` on all 18 changed Python files OK; local `ruff check` on all changed files passed. No prior review threads exist; nothing re-raised. Targeted pytest skipped locally (no env in the worktree; installing the uv dependency set is heavy) — exact-head CI's full suite substitutes.

**Quality-gate summary:** PR-description accuracy verified line-by-line — every claim matches the final state, including "Migration 0037 … does not register new models" (confirmed: the migration only adds the column, backfills the three pre-existing provider-implemented rows — phonely/openai/google — and creates the legacy-route invariant index; no dead GPT-5 seed content survived the final commit). The admin-API registration path works end-to-end: POST with `llm_config` → validated → stored → routed via `benchmark_model` query → per-model client cache and TTFT attribution, all proven by real-DB integration tests that fail without the change. Migration is essential and safe: nullable jsonb column with check, non-destructive, backward-compatible for reads both directions, follows the repo's single-`op.execute` pattern, rollback restores provenance. Auth on the new admin surface is the same `require_coval_admin` + If-Match + `never_shared` as existing routes; no `llm_config` leak into public `/v1/providers` (explicit `ModelInfo` mapping). All call sites migrated to tuple keys; no stale consumers of `make_clients` or the removed `COVAL_LLM_*_AGENT_ID` output. Tests are real-behavior (real Postgres, real factories, real streaming parse, cache-invalidation on registry edits) — none weakened or tautological; one pre-existing test was strengthened.

**Residual risk:** live-platform claims (Coval preserving the `benchmark_model` query on chat/initialization endpoints) rest on the author's smoke calls; pre-migration deploy window leaves admin writes 500ing until 0037 applies — the repo's documented deploy-ahead-of-migrations pattern (0029 color precedent), with rollout order stated in the PR body; downgrade is inherently lossy for admin-registered GPT-5 rows (re-registration required after re-upgrade).

coval-cale:
Closing while we focus on the metrics

PR #663, opened 2026-09-16
Add AssemblyAI universal-3.6-pro streaming STT to the benchmark

### LINKED ISSUE #662: Add AssemblyAI Universal-3.6 Pro streaming (universal-3.6-pro) to the STT benchmark
I'd like to propose adding AssemblyAI's Universal-3.6 Pro streaming model to the STT benchmark, as `universal-3.6-pro` under the existing `assemblyai` provider. Opening this issue first to discuss the approach, per CONTRIBUTING.

### Proposed changes

- Register `universal-3.6-pro` in the existing AssemblyAI v3 streaming provider, wired identically to `universal-3.5-pro` so the two are directly comparable: `speech_model=universal-3-6-pro`, `mode=min_latency`, `end_of_turn_confidence_threshold=1.0`, `ForceEndpoint` at speech end for TTFS. No `format_turns`, keyterms, or prompt.
- Provider tests for the model name, speech-model mapping, ForceEndpoint ordering, and the min_latency preset.
- An Alembic migration registering the model row (`STT / assemblyai / universal-3.6-pro`) with the same feature tags as 3.5-pro, plus a Postgres-backed test for the seeded row.
- No pricing row yet: AssemblyAI has not published a rate for this model. Color left NULL so the site picks one.

### Registered as a "dark" entry

The row is seeded with `collected=False`, `published=False`, `arena_enabled=False` so nothing runs or shows until you flip `collected` on in the admin registry, once the benchmark's AssemblyAI key is confirmed to have access to the model.

### Local verification

I ran the runner CLI against an embedded Postgres with the registry scoped to the three AssemblyAI models, over all 50 `stt-v1` items: 700 results, 0 failures. TTFT and TTFS are a wash against 3.5-pro item by item; mean WER improves from 2.59% to 1.78%, and 3.6-pro was never worse than 3.5-pro on any clip. Full numbers will be in the PR.

Happy to adjust the shape (for example, dropping the migration if you'd rather create the registry row yourselves).

dlange-aai:
Closes #662

Adds AssemblyAI's Universal-3.6 Pro streaming model to the STT benchmark, wired exactly like `universal-3.5-pro` so the two are directly comparable: `speech_model=universal-3-6-pro` on the v3 WebSocket, `mode=min_latency`, `end_of_turn_confidence_threshold=1.0`, and `ForceEndpoint` at speech end so TTFS is measured the same way. No `format_turns`, keyterms, or prompt.

The migration registers the model row (`STT / assemblyai / universal-3.6-pro`) with the same feature tags as 3.5-pro and both `collected` and `published` off. No pricing row: AssemblyAI has not published a rate yet. Color is left NULL so the site picks one.

**After release:** flip `collected` on for `STT / assemblyai / universal-3.6-pro` in the admin registry once the benchmark's AssemblyAI key is confirmed to have access to the model, then `published` when results should show.

**Local verification** (runner CLI against an embedded Postgres, registry scoped to the three AssemblyAI models, all 50 `stt-v1` items, one session each):

| Model | TTFT median | TTFS median | WER mean |
|---|---|---|---|
| universal-3.6-pro | 0.520 s | 0.141 s | 1.78% |
| universal-3.5-pro | 0.520 s | 0.118 s | 2.59% |
| universal-streaming | 0.901 s | excluded | 4.57% |

700 results, 0 failures. Latency is a wash between the Pro models item by item; 3.6-pro was never worse than 3.5-pro on WER for any clip.

Tests: provider tests cover the model name, speech-model mapping, ForceEndpoint ordering, and the min_latency preset; a Postgres-backed test asserts the seeded row's flags and tag set. Upgrade, downgrade, and re-upgrade of the new revision were exercised locally.


seribaymadina (approved):

coval-cale:
Added!

PR #665, opened 2026-09-16
Add a runner-side registry for the S2S caller personas

seribaymadina:
The bank caller personas' labels, descriptions, axis order and dataset ids have no home in the runner. This adds `personas.json` as their single declaration, with a pydantic loader that rejects unknown keys, gapped ordering and a dataset id bound twice. A test pins every binding to `conditions.py` so the file and the condition tables cannot drift. The shared annotated-model base moves to its own module as a prerequisite.

Nothing reads the registry yet. The API and web wiring are follow-ups.


runner/src/coval_bench/scenarios/personas.json
```diff
@@ -0,0 +1,58 @@
+{
+  "_comment": [
+    "The caller personas: what each is called, how it is described to a reader,",
+    "and where its runs are stored. Coval owns what a persona does (noise,",
+    "speakerphone, cell audio); this file owns how we present it and which",
+    "dataset id its conversations land under. Coval ids live in Terraform.",
+    "",
+    "personas: one entry per slug. The slug is an identifier and never changes;",
+    "the label is display text and may. order is the x-axis position, clean first.",
+    "bindings: per family, which personas run and which dataset id each encodes.",
```
seribaymadina, line 10: s2s and llm are seperate families @coval-cale

runner/src/coval_bench/scenarios/personas.py
```diff
@@ -0,0 +1,83 @@
...
+                    raise ValueError(
+                        f"dataset id {binding.dataset_id!r} is bound twice: "
+                        f"{bound_by[binding.dataset_id]} and {family}/{slug}"
+                    )
+                bound_by[binding.dataset_id] = f"{family}/{slug}"
+        return self
+
+
+@functools.cache
+def load_personas() -> PersonaRegistry:
```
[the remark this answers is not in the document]
seribaymadina: Fixed in e83f90ec: `personas` and `bindings` are now typed as `Mapping`, so mutating the cached registry is a mypy error under the `--strict` gate, matching `Scenario.families`.

runner/src/coval_bench/scenarios/personas.json
coval-cale: yewww

coval-cale (approved):

PR #666, opened 2026-09-16
Expose each dataset's caller persona on the by-dataset API

seribaymadina:
Each block of `GET /v1/results/aggregates/by-dataset` gains an optional `persona` field with slug, label, description, axis order and anchor metric, read from the registry #665 adds. Datasets with no binding carry `null`. Numbers and block order are unchanged, and the web ignores the field until its own PR reads it.

Builds on #665; the shared commit disappears once that merges and this is rebased onto main.


coval-cale (approved):
noice

PR #667, opened 2026-09-16
Add the LiveKit worker for the voice-orchestration benchmark

seribaymadina:
LiveKit is the fourth platform after Vapi, Telnyx and Retell, and the first where nobody hosts the agent for us. Vapi, Telnyx and Retell run the process and we patch its config; LiveKit runs nothing, so this package is the process. It lives in the runner package as `coval_bench.livekit_agent` and ships as a third image from the same source, `runner/Dockerfile.livekit`, the way `Dockerfile.api` does for the API.

It reads the pinned stack, tools, prompt and greeting through the existing contract loaders, so nothing is copied. Tool schemas come from `render_livekit_tools` beside the other renderers, verbatim from the contract, and every call goes through a new generic `livekit` codec entry with the simulation id read from the SIP participant attribute at call time. Turn taking is VAD only: Silero 0.3 s plus a fixed 0.5 s endpointing delay for the 800 ms target.

The `livekit` extra is installed only by this Dockerfile and CI, so the runner and API Dockerfiles are untouched. Resolving it did move `openai` 2.41→2.54 in the shared lock, which the runner and API images pick up on their next rebuild; the full suite ran against 2.54 in CI. Prompt, greeting and the Deepgram keyterm list arrive as deployment secrets (`SYSTEM_PROMPT[_B64]`, `FIRST_MESSAGE[_B64]`, `DEEPGRAM_KEYTERMS`, all required at startup) until the dental prompt is committed. CI builds the image; deploying it is a follow-up, since LiveKit Cloud builds from source below Enterprise and only recognises a file named `Dockerfile`.

| | Vapi / Telnyx / Retell | LiveKit |
|---|---|---|
| Who runs the agent process | the vendor | us, this image on LiveKit Cloud |
| Where the config lives | vendor agent object, patched by `platform-assets apply` | the image; a change is a rebuild and redeploy |
| Code in this repo | client + renderer + pins in `platform_assets.py` | renderer + codec entry in the shared files, runtime in `livekit_agent/` |
| Runtime deps | none beyond httpx | `livekit` extra, this Dockerfile and CI only |
| Provider keys | vendor side, or refs handed to the vendor | LiveKit agent secrets |
| Tool wire shape | vendor specific, own codec | generic, existing codec |
| Simulation id on tool calls | vendor header template | read from the SIP participant attribute |
| Runner spec (`AGENTS`, settings) | present | #670 |
| Everything after `/mock` | identical | identical |


runner/src/coval_bench/platform_assets.py
```diff
@@ -490,6 +490,20 @@ def prepare_retell(client: AgentClient, spec: PlatformAgentSpec, dry_run: bool)
     return [f"route:{number}={agent_id}@{RETELL_ROUTE_VERSION}"]
 
 
+def render_livekit_tools(
```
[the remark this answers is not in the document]
seribaymadina: By design for this PR: the runner rows (`PLATFORMS` stub, `livekit-dental` spec, settings fields) are the follow-up PR, as the body says. Until then the agent is deployed with `lk agent deploy` and the Coval agent registered by hand.

runner/src/coval_bench/livekit_agent/contract.py
```diff
@@ -0,0 +1,62 @@
...
+    system_prompt = _text(suite, PROMPT_FILE, "SYSTEM_PROMPT")
+    if not system_prompt:
+        raise ValueError(f"no system prompt: commit {suite}/{PROMPT_FILE} or set SYSTEM_PROMPT")
+    return Contract(
+        suite=suite,
+        stack=load_stack(),
+        tools=tools,
+        system_prompt=system_prompt,
+        first_message=_text(suite, FIRST_MESSAGE_FILE, "FIRST_MESSAGE"),
+        digest=public_contract_sha256(suite),
```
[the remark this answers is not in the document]
seribaymadina: Fixed in 7f09c655: the contract now carries a `prompt_digest` over the resolved prompt and greeting, logged beside the public digest at startup. The public digest stays as is so it remains comparable with the other three platforms.

runner/src/coval_bench/livekit_agent/agent.py
```diff
@@ -0,0 +1,88 @@
...
+async def answer(ctx: JobContext) -> None:
+    ctx.log_context_fields = {"room": ctx.room.name}
+    participant = corr.sip_participant(ctx.room) or await ctx.wait_for_participant()
+    correlation = corr.from_attributes(dict(participant.attributes))
+    logger.info(
+        "call_started",
+        extra={"simulation_id": correlation.simulation_id, "source": correlation.source},
+    )
+    mock = client_from_env()
+    tools = build_tools(CONTRACT.tools, mock.call, correlation)
```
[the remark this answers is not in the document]
seribaymadina: Fixed in 7f09c655: `mock.aclose` is registered as a job shutdown callback, so each session's pool closes when the call ends.

runner/src/coval_bench/livekit_agent/agent.py
```diff
@@ -0,0 +1,100 @@
...
+
+logger.info(
+    "agent_contract",
+    extra={
+        "suite": CONTRACT.suite,
+        "contract_digest": CONTRACT.digest,
+        "prompt_digest": CONTRACT.prompt_digest,
+        "agent_name": AGENT_NAME,
+    },
+)
```
callumreid: **P2 — the `agent_contract` provenance log is dead code; it can never emit.** This module-level `logger.info` runs at import time. At that point no logging handler exists: importing `livekit.agents` installs none (verified: root handlers empty), `coval_bench` configures logging only via an explicit `configure_logging()` call that nothing here makes, and `cli.run_app(server)` — which installs livekit's root handler — runs only after the module body completes. In a livekit `start` deployment the record is emitted in the launcher (no handler) or in a spawned job proc at unpickle time, still before `proc_main` adds the `LogQueueHandler`; with no handler and effective level WARNING, INFO records are dropped. Reproduced: importing the module with placeholder secrets emits nothing. Concrete failure: the prompt arrives only via the `SYSTEM_PROMPT_B64` deployment secret, and `prompt_digest` is recorded nowhere else — not on the mock API, not in the transcript — so a published LiveKit run's actual prompt is unauditable from logs, defeating exactly what this digest was added (commit 7f09c655) to guarantee. Validation misses it because no unit test imports `agent.py`, mypy/ruff can't see runtime log config, and a dropped INFO line produces no error. Smallest sound fix — delete this block and emit per call inside the entrypoint where livekit's handler is active (per-call emission also matches `call_started`'s existing placement), or emit once per job process via `AgentServer(setup_fnc=...)`, which runs after livekit configures logging:

```python
@server.rtc_session(agent_name=AGENT_NAME)
async def answer(ctx: JobContext) -> None:
    ctx.log_context_fields = {"room": ctx.room.name}
    logger.info(
        "agent_contract",
        extra={
            "suite": CONTRACT.suite,
            "contract_digest": CONTRACT.digest,
            "prompt_digest": CONTRACT.prompt_digest,
            "agent_name": AGENT_NAME,
        },
    )
```

runner/src/coval_bench/livekit_agent/agent.py
```diff
@@ -0,0 +1,100 @@
...
+logger = logging.getLogger("coval_bench.livekit_agent")
+
+AGENT_NAME = os.environ.get("AGENT_NAME", "coval-bench-dental")
+SUITE = os.environ.get("SUITE", "dental")
+VAD_MIN_SILENCE_SECONDS = 0.3
+
+
+def keyterms() -> list[str]:
+    raw = os.environ.get("DEEPGRAM_KEYTERMS", "")
+    return [term.strip() for term in raw.split(",") if term.strip()]
```
callumreid: **P2 — Deepgram keyterms default to silently empty, and nothing makes the effective value observable.** `stack.json` pins `keyterms_source` and says dental vocabulary "is recovered through keyterm prompting … generated from the mock fixtures, never hand-maintained" — nova-3 parity is load-bearing for the identity-repair scenario class. This channel reads `DEEPGRAM_KEYTERMS` (not mentioned anywhere in the PR body's secret list), defaults to `[]` (verified: the deepgram plugin omits the keyterm param for an empty list, so STT runs unbiasmed, no error), and no log line or digest records what actually ran — in contrast to the prompt, which gets `prompt_digest` for exactly this override-visibility reason. Concrete failure: the hand deploy sets the documented secrets, `DEEPGRAM_KEYTERMS` is never set because only this code knows the name, and the LiveKit cohort silently under-recognizes dental vocabulary in a published cross-platform comparison. CI misses it: no test covers the empty-env path and deploy-time env is invisible to CI. Non-blocking here only because deploy is a declared follow-up, but it should land before the first cohort. Smallest sound fix: fail loudly like the prompt does, and document the secret in the deploy follow-up:
```suggestion
def keyterms() -> list[str]:
    raw = os.environ.get("DEEPGRAM_KEYTERMS", "")
    if not raw:
        raise ValueError("DEEPGRAM_KEYTERMS is unset; stack.json pins keyterm prompting for nova-3")
    return [term.strip() for term in raw.split(",") if term.strip()]
```
Note `mocktools/keyterms.py::extract_keyterms` already exists as the fixture-derived generator if the deploy lane can populate it.

runner/pyproject.toml
```diff
@@ -57,6 +57,7 @@ google-stt = ["google-cloud-speech>=2.27"]
 google-tts = ["google-cloud-texttospeech>=2.29"]  # >=2.29 for Gemini-TTS streaming
 hume-tts = ["hume>=0.7"]
 hf-parquet = ["pyarrow>=17"]  # build-time only: parquet fallback for unconverted HF datasets
+livekit = ["livekit-agents[deepgram,openai,elevenlabs,silero]~=1.6"]  # the LiveKit agent image only
```
callumreid, line 60: **P3 — body-claim precision: "the runner and API images are unchanged" is not quite true.** The Dockerfiles are untouched, but `openai` is a base dependency (`openai>=1.40`), so the `uv.lock` move 2.41→2.54 that resolving this extra forces is resolved into the runner and API images too on their next rebuild — they silently gain openai 2.54. The body does disclose the shared lock move, so this is wording, not substance: the full suite (API + provider tests included) ran against openai 2.54 in exact-head CI, which is the evidence that makes the bump safe. Worth one clarifying sentence in the body so the next runner/API image rebuild isn't a surprise in its diff.

runner/src/coval_bench/livekit_agent/agent.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def endpointing_delay(contract: Contract) -> float:
+    """The fixed delay after VAD silence that lands on the pinned end-of-turn target."""
+    target = contract.stack.turn_taking.end_of_turn_target_ms / 1000
+    return max(target - VAD_MIN_SILENCE_SECONDS, 0.0)
```
callumreid: **P3 — `endpointing_delay` has no test.** `stack.json` calls turn taking "the layer under test," and the 800 ms → 0.5 s derivation (0.8 − 0.3 VAD) is the PR's headline parity number, yet `test_livekit_agent.py` never pins it — the tests cover contract, correlation, and tools but skip the one value this platform derives on its own. If the target pin ever changes, nothing fails here. Small addition to `test_livekit_agent.py`: `assert endpointing_delay(load_contract("dental")) == 0.5` (plus a sub-0.3-target case for the `max(..., 0.0)` clamp).

callumreid (dismissed):
Reviewing 9fa832d6 against base ca19badd (merge-base with main). Scope verified: the `coval_bench.livekit_agent` package, `Dockerfile.livekit` (analogous to `Dockerfile.api`), the `LIVEKIT = _bare("livekit")` codec entry, `render_livekit_tools` beside the other renderers (not in `PLATFORMS` — deploy genuinely unwired, no half-wiring in `scripts/` or CI beyond an image build), the livekit extra confined to `Dockerfile.livekit` + CI (`test_provider_config` exempts it alongside `hf-parquet`), and VAD turn-taking (Silero 0.3 s, `endpointing_delay` = 800 ms pin − 0.3 = 0.5 s, preemptive generation off — all match the pinned stack). Body claims verified in detail: tool schemas contract-verbatim, simulation id traced SIP-header-attribute → `x-coval-simulation-id` header on mock calls (missing attribute degrades gracefully: call answered, row logged uncorrelated — same as the generic codec), prompt/greeting via `SYSTEM_PROMPT[_B64]`/`FIRST_MESSAGE[_B64]` with a fail-loud guard on the prompt, no secrets committed or logged (gitleaks green), lock delta audited: openai 2.41→2.54 is the only pre-existing package change, zero removals — but see the P3 on the "images unchanged" phrasing. Validation: exact-head CI green (Runner CI run head_sha = 9fa832d6, confirmed via API); locally `uv sync --frozen --dev --extra livekit`, ruff, format, `mypy --strict` (405 files), focused pytest (67 passed incl. all livekit agent tests), module import with placeholder secrets (5 dental tools, delay 0.5, Silero VAD loads), and full livekit-agents 1.8.2 API-surface introspection (`TurnHandlingOptions` TypedDict keys, `raw_schema` → `raw_arguments` binding convention, `add_shutdown_callback`, `rtc_session`).

**REQUEST_CHANGES on one defect inside this PR's own diff:** the provenance log line this PR adds (contract/prompt digests, added to resolve the earlier review round) provably never emits — it runs at module import, before any logging handler exists, and Python drops INFO records with no handler (reproduced locally and traced through livekit's job-proc spawn order in installed source). The prompt-digest audit mechanism for the one platform whose prompt arrives out-of-band is therefore dead code — see the inline P2 at `agent.py:63`. The Deepgram-keyterms P2 is non-blocking (deploy is a declared follow-up) but should land before the first cohort. The two P3s are inline.

Advisory, non-blocking: this PR's merge base predates the recently merged #665/#666/#668; the post-branch main changes touch none of this PR's files (verified: zero overlap), so the merge is textually clean — re-eyeball the combined tree at merge time. Sibling PR #669 (persona cleanup) shares no files or mechanisms with this one; either order merges cleanly.

Residual risk after the fix: AgentSession wiring is type- and API-verified but has no end-to-end voice call in evidence (the deploy lane is the declared follow-up).

coval-cale (approved):

PR #669, opened 2026-09-16
Drop the unused persona_for_dataset lookup

seribaymadina:
After #666 the API resolves a dataset's persona through `binding_for_dataset`, which returns the slug, the persona and the binding. `persona_for_dataset` returned a subset of that and had no consumer left except its own test. Remove it so there is one lookup path, and fold its assertions into the binding test.


callumreid (approved):
Reviewing 62c032ab (single commit) against current main (6556f70f, which contains #665/#666). Scope: remove `persona_for_dataset` (definition, import, `__all__`, its dedicated test) from `scenarios/` and `test_personas.py`. Body claims verified: on origin/main the function has no consumer except its own test (git grep — only `personas.py` def/export and `test_personas.py`); `binding_for_dataset` is the API's live path (`api/schemas.py:308`, from #666) and returns the same slug+persona plus the binding, so one lookup path remains. The assertion fold is genuine, not wholesale deletion: the removed test's `resolved is not None`, `(slug, label) == ("hard", "Hard")`, `DATASET_ID_DENTAL is None`, and `"no-such-dataset" is None` all reappear in `test_binding_for_dataset_carries_the_anchor`, routed through the superset return. Validation: exact-head CI green (Runner CI head_sha confirmed = 62c032ab); locally ruff, format, `mypy --strict` (402 files), and `pytest tests/unit/test_personas.py` (8 passed). No migration, no public API surface affected (the function was exported but unreferenced outside its test). Advisory: sibling PR #667 (LiveKit worker) shares no files or mechanisms with this change and branched before `scenarios/personas.py` existed — zero interaction, either order merges cleanly. Ready.

PR #670, opened 2026-09-16
Add the LiveKit arm to the platform agent table

seribaymadina:
LiveKit joins the platform agent table as `livekit-dental`, so `platform-assets register`, `launch` and the `--agent` choices know the fourth arm the way they know Vapi, Telnyx and Retell. Its dial target and dispatch name come from two new secrets, `LIVEKIT_DENTAL_DIAL_TARGET` and `LIVEKIT_DENTAL_AGENT_NAME`; the Coval agent it registers is the same body as the other three, only the `sip:` target differs.

What it does not get is a vendor client. LiveKit has no agent object to read or patch: the agent's configuration is the deployed image, and a change is a rebuild. So `livekit` is listed as code-managed rather than given a `Platform` row. `plan` and `apply` refuse it with that explanation, `launch` skips the vendor drift check for it, and the spec's vendor API key becomes optional, which is the one model change. The runner never holds LiveKit credentials; those belong to the deploy lane.

The refusal is not the last word on drift. No workflow checks any platform for drift today; the daily cohorts dial the agents without passing through `launch`. A scheduled `plan` over every spec, visible-red like the arena key-parity guardrail, is the follow-up, and LiveKit joins it once the deploy lane stamps the contract digest on the agent for a read step to compare.

The agent itself, its codec entry and tool renderer are #667. This PR is independent of it and touches only the table, the guard and their tests.


runner/src/coval_bench/platform_assets.py
```diff
@@ -988,9 +1027,7 @@ def assets_launch(
     """Launch one Coval run against this variant, once both the platform and Coval agent match."""
     spec = spec_for(agent_key)
     install_fixture_providers(Settings())
-    platform = platform_for(spec)
-    with platform.client(spec.api_key.resolve(), api_base or platform.api_base) as client:
-        pending = drift(client, spec, mock_base_url)
+    pending = vendor_drift(spec, mock_base_url, api_base)
```
callumreid, line 1057: **P3 — `--mock-base-url` is still required on the LiveKit arm but never used.** `platform-assets launch --agent livekit-dental` requires `--mock-base-url` (`required=True` on the option, with the `MOCK_TOOLS_BASE_URL` envvar fallback), but on the LiveKit arm the value is now never consumed — `vendor_drift` returns `[]` for code-managed platforms before touching it, and neither `register` nor `launch_body` reads it. An operator or script must supply a plausible URL that the command ignores, misrepresenting the CLI contract for the fourth arm. Existing tests cover `vendor_drift` and `coval_agent_body` directly, not the CLI surface, so nothing catches it. Smallest sound correction: make the option optional for code-managed arms (resolve the spec first and pass `required` conditionally, or validate `mock_base_url` only when `spec.platform not in CODE_MANAGED`), or at minimum note in the option help that it is unused for code-managed arms. Non-blocking — the other three arms genuinely need it.

callumreid (approved):
**Verdict: READY — approved**, one non-blocking P3 inline.

Reviewed the full diff at head `8b2c94ea` (merge-base `6556f70`; 2 files, +78/−8) against freshly fetched `origin/main`, reading all of `platform_assets.py` (1054 lines) and its test file at head, every in-repo consumer, and the CLI surface.

Verified clean: `api_key: SecretRef | None` is guarded at every consumer (`plan`/`apply` refuse code-managed platforms before key resolution; `_vendor_key` raises a named `SyncError` on `None`; `vendor_drift` short-circuits on `CODE_MANAGED`; `smoke`/`register`/bodies never touch it); vendor-drift semantics for the three vendor arms are preserved exactly; the new arm follows the table's structure (fields, `MOCK_TOOLS_SECRET`, purpose strings); no dial-target parsing needed for LiveKit, matching Vapi; tests assert behavior (known-set message names four arms, code-managed refusal, drift gate returns `[]`, LiveKit `sip:` body) and fail without the change (`spec_for("livekit-dental")` raises on main).

Validation: 85 unit tests passed locally with proven worktree resolution (environment created outside the worktree; `coval_bench.__file__` verified to point into it); exact-head CI all green (ruff, ruff format, mypy --strict, full pytest, gitleaks, docker builds, CodeQL). PR body accurate — including the disclosed #667 codec dependency; the `--agent livekit-dental` choice on `smoke` fails with a raw `KeyError` until #667 lands (sequencing nuance the body already discloses; advisory). Residual risk: low.

runner/src/coval_bench/platform_assets.py
seribaymadina: Fixed in eed4e471. `--mock-base-url` is no longer required at the click level; each command resolves it through `mock_base_url_for`: plan, apply and smoke always need it, launch only for vendor-managed arms, so a LiveKit launch no longer demands a value it never reads. The option help says so. Covered by a helper test and a CliRunner test of the launch surface for both arm kinds.

runner/src/coval_bench/platform_assets.py
```diff
@@ -581,9 +583,32 @@ def prepare_retell(client: AgentClient, spec: PlatformAgentSpec, dry_run: bool)
             purpose="the sip:+E164@sip.retellai.com URI Coval dials; names the imported number",
         ),
     ),
+    PlatformAgentSpec(
+        key="livekit-dental",
+        platform="livekit",
```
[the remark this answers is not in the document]
seribaymadina: The codec ships with #667, which merges before this PR, so the arm and its codec land in order. As of eed4e471 smoke also catches the missing-codec case and raises a named click error instead of a raw KeyError, so running it out of order fails with a message rather than a traceback.

PR #672, opened 2026-09-17
[BENCH-940] Index capture times for results pagination

kdmelon:
## Summary

Cross-run Results API v2 requests scan and sort their capture-time window. Add a nonpartial `(captured_at DESC)` index so pages can start at recent observations. The evaluation-ID tiebreaker still requires sorting within equal capture timestamps; an observation UUID key would not help it.

## Review guide

Start with `runner/src/coval_bench/db/migrations/versions/20260917_0038_results_capture_index.py`: an Alembic autocommit block drops any same-name index and recreates it concurrently, recovering from interrupted attempts before the revision is recorded. The focused migration tests exercise an actual canceled concurrent build, a valid preexisting index, upgrade/downgrade, and preservation of rows and other indexes.

## Testing

- PostgreSQL migration tests: **3 passed**, including real canceled-build recovery, valid-index retry and downgrade.
- Real PostgreSQL migrations + RunWriter + HTTP smoke passed with historical null IDs unchanged, components and cursor pages; public OpenAPI unchanged. Repeated HTTP requests also passed with custom and generic prepared plans.
- Same 100,000-observation / 400,000-evaluation dataset: capture-only index **958,464 bytes**, versus **4,079,616 bytes** with the unused UUID key; matching rows and the same ordered-scan/incremental-sort access path for normal pages.
- First-page SQL with the original compatible query: **0.63 ms** for five rows and **27.7 ms** for 1,000 rows, versus the earlier unindexed **3.05 s / 3.29 s**.
- Changed-file Ruff/format and diff checks passed; independent Sol re-review passed. Strict local mypy remains blocked by an existing dependency shim shadowing `typing_extensions`.

## Rollout and limitations

Merge and apply this migration before [API optimization #673](https://github.com/coval-ai/benchmarks/pull/673). Retrying an unrecorded migration rebuilds even a valid same-name index; run migrations serially. Concurrent creation preserves ordinary writes but still consumes database resources.

A single tied timestamp covering all 400,000 evaluations still requires scanning and sorting the cohort; the revised compatible API took about **4.94 seconds** in that synthetic case. Local timings do not establish production latency. No production migration or deployment was performed.

Fixes BENCH-940


borgesius (changes requested):
The access path is directionally right, but I do not think we should build this exact index in production yet. The API orders by `(o.captured_at, e.id)`, where `e.id` is the metric-evaluation ID; the second column here is the observation ID, so PostgreSQL only treats `captured_at` as the presorted key. On a disposable PostgreSQL 16 database with 100,000 observations and 400,000 evaluations, `(captured_at DESC)` produced the same index-scan + incremental-sort plan while occupying about 2.2 MB versus 12 MB for `(captured_at DESC, id DESC)`. Please drop the unused UUID key unless another measured consumer needs it. Also make the concurrent migration restart-safe, or document the exact invalid-index recovery: an interrupted `CREATE INDEX CONCURRENTLY` can leave the named index INVALID while Alembic remains on 0037, and a plain retry then fails because the relation already exists. One plan limitation to record in the evidence: a deliberately tied capture-time cohort still scanned/sorted 400,000 evaluations and took about 272 ms, since the evaluation-ID tiebreaker cannot be satisfied by this observation index.

borgesius (approved):
The revised index is the right access path: the unused observation UUID key is gone, interrupted concurrent builds recover cleanly, and the focused PostgreSQL migration suite passes. Independent PostgreSQL 16 planning also confirms that this index enables the bounded ordered scan used by the companion query. Rollout ordering matters: apply and verify this migration before releasing #673; merging it alone does not establish the required production index.

PR #673, opened 2026-09-17
[BENCH-938] Optimize cross-run results queries

kdmelon:
## Summary

Speed up cross-run Results API v2 pagination with an indexable capture-time cursor bound and a direct metric catalog join on `e.metric_id`. The NOT NULL and validated catalog foreign-key guarantees from merged #675 make the legacy null-ID resolver and its lateral lookup unnecessary.

Response fields, eligibility filters, visibility, ordering and cursor format remain unchanged.

## Review guide

Start with `runner/src/coval_bench/api/routers/results_v2.py`: the direct catalog join uses the enforced metric ID, every eligibility filter remains before the page limit, and the strict capture-time/evaluation-ID cursor is retained. Tests cover backfilled historical rows and fresh writes after actual NOT NULL enforcement, pagination and canonical metric filtering, plus equal-time ordering across observations and ineligible newer rows.

## Testing

- Focused v1/v2 API suite: **42 passed** (18 v2, 24 v1).
- Real migration smoke passed: pre-0036 historical rows → hydrate all three metric-ID tables → actual 0039 → fresh RunWriter writes → HTTP pagination/components. Verified NOT NULL and validated foreign keys on all three tables; public OpenAPI unchanged.
- Repeated HTTP smoke passed on 400,000 evaluations with enforced IDs: 14 first-page and 14 deep-cursor requests using custom and generic prepared plans. Local median HTTP times were 1.3 ms for five rows and 14.0 ms for a deep 1,000-row page.
- Changed-file Ruff/format/compile/diff checks and independent Sol review passed. Scoped strict mypy passed for both changed files with `--follow-imports=silent --ignore-missing-imports` to isolate the existing dependency environment.

## Rollout and limitations

Apply capture-time index migration 0038 ([#672](https://github.com/coval-ai/benchmarks/pull/672)) and metric-ID enforcement migration 0039 ([#675](https://github.com/coval-ai/benchmarks/pull/675)) before deploying this API version. Both PRs are merged; production application of 0039 is operator-confirmed. Older nullable schemas are no longer supported by this query.

The observation index orders capture times; large equal-time cohorts still need evaluation-ID sorting. Local synthetic performance tests do not establish production latency. This PR task has not deployed the API or changed the production database.

Fixes BENCH-938


ben-coval (approved):

borgesius (changes requested):
The cursor bound is a substantial improvement: on a local 100,000-observation/400,000-evaluation PostgreSQL shape, a deep page changed from about 194 ms and roughly 204k shared-buffer accesses to 0.08 ms and 19 buffers once `o.captured_at <= anchor_time` became an index condition. The direct metric join is premature, though. #656 explicitly staged historical NULL compatibility until BENCH-917 enforces `metric_id NOT NULL`, with compatibility removal afterward under BENCH-918. That enforcement has not landed; the column is still nullable and this inner join silently drops any missed historical row. A production backfill assertion in the PR body does not protect every target environment or make future drift fail closed. Please land and verify the enforcement/validated constraint first, or preserve a bounded NULL fallback here. This also must deploy only after the corrected #672 migration is applied.

borgesius (approved):
The updated query preserves historical NULL metric-ID fallback, unknown-code behavior, eligibility filtering before LIMIT, and stable same-timestamp pagination. Independent PostgreSQL 16 comparison on 100k observations / 400k evaluations reduced a representative deep page from about 205k shared-buffer accesses and 89 ms to about 1k buffers and 5.3 ms. Rollout ordering matters: do not release this API change until #672 has merged and its migration has actually been applied and verified.

PR #674, opened 2026-09-17
[BENCH-939] Add StepFun StepAudio TTS and ASR providers

coval-cale:
StepFun serves StepAudio TTS over a WebSocket that opens a session before any text is sent, so t0 is the text submission; sentence mode returns 24 kHz PCM as base64 deltas. The STT provider speaks StepFun's Realtime-shaped ASR socket with no server VAD: paced base64 appends, a manual commit, and a cumulative hypothesis whose first words arrive only in the correctable stash field.

Model rows are added on /admin after merge; no migration.


seribaymadina (approved):

runner/src/coval_bench/providers/stt/stepfun.py
```diff
@@ -0,0 +1,261 @@
...
+                    if transcript:
+                        set_first_token(result, transcript, now=now)
+                        add_partial_transcript(result, transcript)
+                        completed_by_item[item_id] = transcript
+                    if result.audio_start_time is not None:
+                        last_final_time = now
+                    if event.get("reason", "commit") == "commit":
+                        saw_commit = True
+                        break
+                    continue
```
callumreid, line 229: **P3 — non-commit completion path is untested.** The `_COMPLETED` handler accumulates per-item finals (`completed_by_item`) and continues on a non-commit `reason`, i.e. it is written to handle server-side segment splits — but no test exercises that path, and its `commit` default has a real truncation hazard: if StepFun's ASR session ever emits a segment completion without a `reason` field, `event.get("reason", "commit") == "commit"` treats it as the manual commit, the loop breaks at the first segment, and the result reports a truncated transcript as a clean success (deltas for later segments are never read). The sibling this provider mirrors, Nari, documents exactly this failure shape (36 s server-side split) and pins it with `fixtures/nari/events-split.json` + `test_nari_joins_auto_split_items_in_order`. The happy-path and error tests here don't cover it, so nothing would catch a wrong default. Smallest sound fix — mirror the Nari split test:

```json
[
  {"type": "session.updated"},
  {"type": "conversation.item.input_audio_transcription.delta", "item_id": "item_001", "text": "hello world ", "stash": "how"},
  {"type": "conversation.item.input_audio_transcription.completed", "item_id": "item_001", "transcript": "hello world how are you", "reason": "max_duration"},
  {"type": "conversation.item.input_audio_transcription.delta", "item_id": "item_002", "text": "goodbye", "stash": ""},
  {"type": "conversation.item.input_audio_transcription.completed", "item_id": "item_002", "transcript": "goodbye", "reason": "commit"}
]
```

asserting both segments are joined in `complete_transcript`, `saw_commit` only fires on the commit final, and the loop does not break on the first completed.

callumreid (approved):
Verdict: APPROVE — one non-blocking P3 inline.

Reviewed: base=origin/main (b90b03d4, freshly fetched), exact head 7b585990b2dba5824026a277605515d585a48035, scope = 13 files: new STT and TTS StepFun providers plus tests/fixtures, Settings field, both provider registries, PROVIDER_ENV, .env.example, methodology doc. Instruction sources: benchmarks README/CONTRIBUTING, runner provider patterns (nari STT, minimax TTS siblings), orchestrator, arena pairing/health code.

Validation evidence: `python3 -m py_compile` on both providers and both test files passed. Exact-head CI green: Runner CI (9m39s) ran ruff check, ruff format --check, mypy --strict, pytest with the new tests, pip-licenses, gitleaks, and both Docker builds against this single-commit head; CodeQL and the Switchboard gate also passed. Local pytest/ruff/mypy skipped (no venv present; installing is out of scope for this run).

Quality-gate summary: both providers reuse the shared contracts — STTProvider/TTSProvider, paced_chunks, transcript utils, finalize_tts_result — and mirror the closest siblings structurally (nari STT, minimax TTS); registration is complete (Settings, .env.example, STT/TTS registries, PROVIDER_ENV, admin module-listing validation); keys are SecretStr end-to-end, never logged or put in URLs, and unrotated placeholders are null+warned by the Settings validator; every failure path lands in result.error, including the shared silent-close guards on both sides; t0/TTFA semantics match the methodology WS-cohort row. PR description matches the diff.

Residual risk: wire-protocol fidelity (event names, reason semantics, session caps) is verified against the author's fixtures, not a live StepFun call. Confidence: High.

Post-merge note: this PR merged while the review was being finalized; this review confirms the merged head 7b585990b2dba5824026a277605515d585a48035. Cross-checked against benchmark-infra#207 (also merged): the STEPFUN_API_KEY env var, the settings field, both registry slugs, and the secret name line up exactly across the two diffs.

PR #675, opened 2026-09-17
[BENCH-941] Enforce non-null normalized metric IDs

kdmelon:
## Summary

Targets `main` after #672 merged. This diff contains only the NOT NULL enforcement migration and supporting tests.

After the metric-ID backfill, enforce non-null catalog identities on `metric_evaluations`, `dashboard_metric_values`, and `metric_values_by_bucket`. Validate their existing catalog foreign keys. Preserve metric IDs, evaluation payloads, code-based retry keys, and identity-populating triggers.

## Review guide

Start with `runner/src/coval_bench/db/migrations/versions/20260917_0039_normalized_metric_id_not_null.py`: committed staging checks let PostgreSQL validate existing rows without retaining the exclusive locks needed for the short DDL steps. All validations finish before any column becomes NOT NULL; interrupted attempts can resume. Migration tests cover the failure and recovery paths.

## Testing

- Final PostgreSQL index/migration/backfill compatibility suite against #672 at `3f321a61`: **34 passed**, including all-table failure, orphan FK, partial interruption, and lock-boundary checks.
- Existing writer/projection/metric-ID checks: 49 passed initially; two historical fixture failures corrected, then all four projection tests passed.
- Real runtime smoke: historical upgrade and resumed backfill, NOT NULL enforcement, fresh RunWriter writes, source/hourly/summary generation, and HTTP v1/v2 responses passed.
- Changed-file Ruff, formatting, strict mypy, and diff checks passed.
- Independent Sol review passed. All disposable PostgreSQL processes stopped.

## Rollout

Apply the preceding capture-time index migration from #672 first. Before applying this migration in each environment, run the existing metric-ID preflight and require `completed`, `verification_complete: true`, zero pending/unknown/mismatched IDs, and no errors. This migration performs no backfill and refuses to complete if null or orphaned IDs remain.

From `runner/` with the target `DATABASE_URL` configured:

```sh
./.venv/bin/python -m alembic -c alembic.ini upgrade 20260917_0039
```

Apply before deploying #673, which removes the results API's null-ID fallback. Each lock acquisition is bounded; validation scans still consume database resources. If an attempt fails, committed staging checks may remain while Alembic stays at the preceding revision; resolve the cause and rerun the same upgrade. Downgrade restores nullable columns while retaining validated foreign keys and all data. No production migration was run by this PR task.

Fixes BENCH-941
Refs BENCH-917


ben-coval (approved):
Reviewed the migration and its tests. The NOT VALID check / VALIDATE / SET NOT NULL / drop-check ordering is correct for Postgres 16's scan-skip path, every step is guarded so an interrupted attempt resumes cleanly at 0038, the FK constraint names match those created in 0036, and the test suite covers the failure, orphan-FK, partial-interruption, and lock-release paths. Single alembic head. CI is green.

Non-blocking notes:

- The null-hydration branch in `writer.py` (around line 478) becomes dead once this lands. Harmless, and #673 removes it.
- `backfill_normalized_metric_ids.py:100` creates a partial index `WHERE metric_id IS NULL` that will always be empty afterwards. Still valid, just no longer useful.
- Pinning the fixtures in `test_backfill_normalized_metric_ids.py` and `test_dashboard_metric_values.py` to specific revisions is the right call for fixtures that seed nullable-stage data. `test_normalized_db_writer` still migrates to head, so head coverage is intact.

borgesius (approved):
The staged NOT VALID proof constraints, committed validation steps, bounded SET NOT NULL locks, and resumable cleanup are a solid migration shape. I reran the focused PostgreSQL migration/backfill/projection/index suites locally: 38 passed. One rollout-description correction: #673 current head preserves the historical NULL metric-ID fallback, so this migration is no longer a prerequisite for that API change; #672 applied capture-time index remains the actual #673 deployment dependency. Please update the PR metadata so the train does not imply otherwise.

callumreid (approved):
**Verdict: READY** — zero findings. Posted post-merge (merged 20:48 UTC while this first review was in flight; the head is unchanged and is what landed). The migration is the correct second half of an already-shipped two-stage design: 0036 (on main) added the nullable `metric_id` columns with NOT VALID FKs, identity triggers, and a resumable backfill CLI; 0039 only enforces — backfill-first-then-enforce, the acceptable pattern.

Verified at `d32dde1b`: single-head revision chain (39 versions, zero branch points, `down_revision` attaches to main's BENCH-940 tip); every statement independently committed with existence pre-checks, so each crash point resumes idempotently — including after the first `SET NOT NULL`, simulated by test via an `after_cursor_execute` hook; `NOT VALID` adds and `VALIDATE CONSTRAINT` keep ACCESS EXCLUSIVE windows catalog-only and bounded by a session-level 10s `lock_timeout` (saved and restored in `finally`); `SET NOT NULL` skips the table scan via the validated CHECK proof constraint; no writer path can produce NULL after enforcement (all inserts/updates resolve the catalog id or are backstopped by the 0036 BEFORE-INSERT triggers; `session_replication_role` appears nowhere under `runner/src`); downgrade restores nullable columns while retaining validated FKs and data, exercised by test with payload equality across the round trip; the pinned fixtures in the two modified test files are required accompaniments with accurate comments, and the unmodified tests in the same files still walk to head.

Validation: exact-head CI green at this SHA (Runner CI 9m23s including the full pytest suite against embedded Postgres — run `head_sha` confirmed equal to the reviewed SHA — plus Analyze Python, CodeQL, Plan checks, Switchboard Gate). Local DB tests skipped (no PostgreSQL binaries on this box); `py_compile` on the 4 changed files clean; the migration chain parsed programmatically. PR description accurate line by line, including the rollout and rollback notes.

Residual risk: low. Advisory cross-PR note: #673 (removes the results API's null-ID fallback) must not serve traffic from an environment still pre-0039 — worth an `alembic_version` check when that ships.

PR #678, opened 2026-09-17
[BENCH-944] Add time to first sentence to the LLM benchmark

coval-cale:
TTFT measures when a model starts talking to the proxy. A voice pipeline cannot speak until the first sentence is complete, so the caller feels time to first sentence. The accumulator stamps the first terminal punctuation followed by whitespace, guarding against spelled-out initialisms like A.M., and records the sentence's length so a slow value can be read as slow tokens or a long sentence. Turns with no boundary fall back to total time; tool-only turns are null. The fetch writes the per-conversation mean as TimeToFirstSentence beside TTFT on the bank dataset.


seribaymadina:
Clock arithmetic looks correct, but three issues affect fairness (`3b0c164`):

- **Boundary detection:** `ACH.` is ignored while `ach.` counts; `Dr. Smith` incorrectly splits at `Dr.`. Fix acronym/abbreviation handling.
- **Backdating:** punctuation at 0.1 s, confirmed by whitespace at 1.1 s, reports 0.1 s. For speech readiness, record when the boundary is confirmed.
- **Fallback:** unpunctuated text is timed to stream closure, including trailing protocol overhead. Use the completion's finish signal.

Use one shared sentence detector and add regression tests for these cases. Label this upstream LLM timing: it excludes tool-only requests and audio delivery, so it is not full caller latency.

seribaymadina (changes requested):
felt like the comments i  left could affect the provider ranking, feel free to push back tho

PR #679, opened 2026-09-17
[BENCH-945] Make normalized benchmark writes recoverable

kdmelon:
## Summary

- Add opt-in fail-closed normalized capture, with GCS and PostgreSQL privilege preflight before provider or import work.
- Persist immutable run manifests, frozen envelopes, exact result-ID allocations, receipts, seals, and finalization receipts for replayable recovery.
- Make S2S and LLM imports restart-stable and concurrency-safe while preserving genuine provider failures.
- Add bounded `coval-bench db capture status/recover` operator workflows, including abandoned unsealed-run recovery.

## Review guide

1. Start with `runner/src/coval_bench/runner/capture.py` and `runner/src/coval_bench/runner/normalized.py` for the durable capture contract and replay invariants.
2. Review `runner/src/coval_bench/db/writer.py` for exact/idempotent database replay, privilege preflight, and transactional finalization.
3. Review `runner/src/coval_bench/runner/orchestrator.py` and `runner/src/coval_bench/s2s/fetch_v2v.py` for restart-stable integration, then `runner/src/coval_bench/runner/capture_cli.py` for bounded operator recovery.

## Testing

- `runner/.venv/bin/pytest -q runner/tests/unit/test_config.py runner/tests/unit/test_observation_artifacts.py runner/tests/unit/test_normalized_dual_write.py runner/tests/unit/test_capture_recovery.py runner/tests/unit/test_orchestrator.py runner/tests/unit/test_s2s_fetch.py` — 236 passed.
- Ruff on all changed Python and test files — passed.
- Strict mypy on the nine changed source modules — passed.
- Targeted strict mypy on the two repaired test files — passed.
- Focused capture conflict/transport logging tests — 2 passed; targeted Ruff, format, and strict mypy passed.
- `git diff --check` and staged `git diff --cached --check` — passed.
- `runner/.venv/bin/coval-bench db capture --help` — passed.
- Read-only production Cloud SQL checks confirmed the required tables and runner-role table/sequence privileges. No production writes were performed.

## Known limitations / rollout

- Required mode is disabled by default; this PR does not deploy or enable it.
- Disposable PostgreSQL concurrency coverage could not run locally because PostgreSQL binaries were absent and Docker was unavailable.
- No real GCS write/collision integration was run against an authorized bucket.

Fixes BENCH-945


runner/src/coval_bench/runner/normalized.py
```diff
@@ -61,6 +413,181 @@ def _inputs(metric: str, artifacts: dict[Any, Any], benchmark: Any) -> list[Metr
...
+                ]
+                await writer.complete_metric_evaluation(
+                    evaluation.id,
+                    values=values,
+                    finished_at=frozen.finished_at,
+                    validate_contract=False,
+                )
+        await asyncio.to_thread(upload_receipt, storage_client, bucket, envelope)
+        return CaptureOutcome.COMPLETED
+    except ValueError:
```
ben-coval: `ValueError` is very wide here. Pydantic's `ValidationError` subclasses it, and so do `ObservationArtifactType(...)`, `MetricExecutor(...)`, the frozen-artifact digest check, and `upload_immutable_object`'s verification failure. A corrupt envelope, a bad enum value in frozen data, and a genuine concurrent-replay conflict all come back as `CaptureOutcome.CONFLICT` with the exception discarded.

Nothing is lost — the envelope stays durable and a later run after a fix still replays — but the operator-facing story dead-ends. `orchestrator.py:779` appends the outcome with no logging, so a required-mode run goes PARTIAL with `normalized capture pending`, `coval-bench db capture recover` reports `{"outcomes": {"conflict": 1}}`, and there is no way to learn the cause short of reproducing it. That cuts against the backlog-visibility and conflict-detection goals in BENCH-945.

A `logger.warning(..., exc_info=True)` in both this handler and the transport handler on line 583 would be enough; the code right below at `orchestrator.py:783` already logs that way for the legacy path.

ben-coval (approved):
Approving. The durability model holds up: the envelope is the acknowledgement boundary, replay is keyed off frozen payloads rather than current mappers, and the exact-replay writes are genuinely idempotent under concurrency. Required mode is opt-in and config validation refuses to enable it without dual-write and a bucket, so the blast radius of landing this is small.

Things I checked and was happy with:

- `record_results_exact` takes an identity-scoped advisory lock and re-verifies the full row set after insert, so a concurrent identical replay collapses to one row and a differing immutable field is a conflict rather than a silent accept. `test_exact_capture_replay_serializes_and_promotes_pending_run` covers exactly that against a real Postgres, and CI runs pytest-postgresql on loopback, so the caveat in the PR description about local Postgres being unavailable does not leave this path unverified.
- `upload_immutable_object` only accepts a precondition conflict after verifying the stored bytes, size, content type and digest, so the `except ValueError` adoption paths in `upload_import_run_claim` and `upload_legacy_result_allocation` are reacting to a verified exact collision rather than any upload failure.
- `mark_run_capture_pending` refuses to relabel a FAILED run, and `finish_run_exact`'s recovery promotion requires the stored error to be exactly `normalized capture pending`, so a genuine provider partial cannot be promoted by recovery.
- The unguarded `ENQUEUE_RUN_BUCKET_SQL` in `mark_run_capture_pending` is reachable only under `required_capture`, which has already run `preflight_required_capture_schema` and asserted `dashboard_source_refreshes` exists. Its two siblings guard `UndefinedTable` anyway; adding the same guard would make the three consistent, but it is not a live failure.

One non-blocking comment inline, plus two coverage notes below. None of it needs to hold up the merge.

Against the BENCH-945 acceptance list, I could not find coverage for publication failure (the dashboard refresh enqueue path) or for a partially-written set of evaluations during replay. Everything else on that list maps onto a named test.

PR #680, opened 2026-09-18
Report per-provider item outcomes in the run summary

### LINKED ISSUE #681: Report per-provider item outcomes in the run summary
Hi! I noticed that when a run finishes, `RunSummary` reports how many result rows succeeded and failed across every provider together, but nothing reports, per provider model, how many items it was asked for and how many it actually delivered. I think that's a really useful number when deciding whether to trust a provider in production: not just how fast it is when it works, but how often it works.

Right now the closest thing is `_dead_providers`, which only fires when a model fails every single item. And since the results tables and API filter to successful rows before aggregating, the failure count isn't recoverable downstream. It has to be counted in the runner.

## Proposal

Add a per-provider breakdown to `RunSummary` in `runner/src/coval_bench/runner/orchestrator.py`, computed once when the run finishes from the results already in memory. It shows up in the JSON line that `coval-bench run` prints and on the `benchmark_run_finished` log event. No new table, migration, or endpoint.

Each entry has three counts:

- `succeeded`: items the provider delivered
- `provider_failures`: items it didn't (5xx, timeouts, closed the stream without audio, and so on)
- `key_failures`: items that failed because of *our* key (auth, out of credit, rate limit), reported separately so they don't count against the provider

Items are counted from one metric per benchmark that is written for every item and fails only when the provider call failed: TTFA for TTS and AudioToFinal for STT. Failures are sorted with the existing `classify_failure` from `arena/provider_health.py`.

Example entry:

```json
{"benchmark": "TTS", "provider": "example", "model": "example-v1", "succeeded": 27, "provider_failures": 2, "key_failures": 1}
```

That reads as 27 of 29 delivered, or 93.1% reliability, with the one key failure not held against the provider. Down the line it would be nice to surface this on benchmarks.coval.ai as a reliability percentage next to each provider's latency numbers, but that's beyond this change.

Draft PR: #680

2 things worth noting:
- First, an item is judged after the runner's retry helper has had its attempts (up to three on timeouts and connection errors), so this counts delivered items, not individual requests. A provider that drops two of every three connections but always recovers within the retry budget would still read as fully reliable. Counting every attempt would be more honest about flakiness, but it needs the retry helper to report attempts, so I've left it for a follow-up.

- Second, this only reports the counts for a single run, in that run's output. Nothing saves them, so there's no way yet to ask something like "how often did a provider fail over the last week?" That would need the counts stored in the database with a query or endpoint on top, which felt like a bigger, separate change. Happy to take that on as a follow-up if it would be useful.

estrindavid:
Closes #681.

A run reports how many result rows succeeded and failed across all providers, but nothing reports, per provider model, how many items it was asked for and how many it delivered. That's the number that says how often a provider works, not just how fast it is when it does.

## What this adds

- A `reliability` list on `RunSummary`, one entry per `(benchmark, provider, model)`.
- It appears in the JSON line `coval-bench run` prints and on the `benchmark_run_finished` log event.
- No new table, migration or endpoint.

Each entry has three counts that partition the items attempted:

- `succeeded`: items the provider delivered
- `provider_failures`: items it didn't (5xx, timeouts, closed the stream without audio, and so on)
- `key_failures`: items that failed because of *our* key (auth, out of credit, rate limit), reported separately so they don't count against the provider

So a provider's reliability is `succeeded / (succeeded + provider_failures)`.

```json
{"benchmark": "TTS", "provider": "example", "model": "example-v1", "succeeded": 27, "provider_failures": 2, "key_failures": 1}
```

## How items are counted

- One row per item, from the metric each benchmark always writes: TTFA for TTS, AudioToFinal for STT. The other metrics are either excluded for some models or only written on success.
- Failures are sorted with the arena's existing `classify_failure`: auth, credit and rate limit are key failures, everything else counts against the provider.
- An item is judged after the retry helper has had its attempts (up to three on timeouts and connection errors), so this counts delivered items, not individual requests. A provider that fails two of every three connections but always succeeds within the retry budget still reads as fully reliable.

Left as follow-ups: counting every attempt rather than every item (needs the retry helper to report attempts), persisting the counts across runs, moving `classify_failure` out of `arena/`, and threading `TTSResult.status_code` through to the result row.

## Testing

- Four unit tests beside the `_dead_providers` ones, plus `test_full_failure` now asserts the breakdown on the summary and on the `benchmark_run_finished` event. 124 passed across the orchestrator, runner, suite, telemetry and logging files.
- Ruff, format and strict mypy clean on the changed files. The DB-backed tests need PostgreSQL 15+ and could not run locally on 14.


estrindavid:
## Results from testing

Ran ~20,000 single-attempt STT sessions over 39 hours to check how common provider errors actually are! Free-tier keys, one laptop, 5 providers, 11 models.

| Provider | Model | Tests | Errors | Error rate |
|---|---|---|---|---|
| deepgram | nova-3 | 4027 | 0 | 0.0% |
| assemblyai | universal-streaming | 961 | 0 | 0.0% |
| gladia | solaria-1 | 179 | 0 | 0.0% |
| speechmatics | enhanced | 51 | 0 | 0.0% |
| speechmatics | default | 49 | 0 | 0.0% |
| speechmatics | linden-1 | 49 | 0 | 0.0% |
| deepgram | flux-general-multi | 4527 | 1 | 0.0% |
| deepgram | flux-general-en | 4563 | 4 | 0.1% |
| assemblyai | universal-3.5-pro | 960 | 1 | 0.1% |
| elevenlabs | scribe_v2_realtime | 549 | 1 | 0.2% |
| deepgram | nova-2 | 4078 | 9 | 0.2% |

**Findings**

- Genuine provider errors are rare: 0–0.2% across every model, and almost all are one-offs.
- One reproducible failure: nova-2 hangs on a specific 2.65s dataset clip — emits a partial, ignores `Finalize`, never sends a final until Deepgram closes the idle socket. 18/18 failures across six occasions and twelve replays. Same clip passes on nova-3, both Flux models, and AssemblyAI 20+ times. Roughly 1 in 900 clips trigger it, so the prod runner likely hits this every couple of days and records it as "no AudioToFinal produced" with no cause.

**So what?**

Provider reliability looks small, but since I was using free tokens and ran all these tests locally, this may not be enough information to make any final conclusions

coval-cale (approved):

